PR #44, opened 2016-03-23
bracket constructor syntax for set does not exist in python < 2.7

bmaranville:
We want to use uncertainties with a package that will sometimes be installed on python2.6.  This was the only line of code that prevented compiling and installing the uncertainties package.


lebigot:
Thanks. Fixed in version da6fd6b.

Issue #61, opened 2016-09-16
Continious Integration


rth:
Travis-CI was disabled in August for this repo. Should we add it back? The python versions that are tested in `.travis.yml` also might need to be updated.

Also I'm not sure if it would be useful to add CI tests on windows with [AppvVeyor](https://www.appveyor.com/)? In which case I can add the setup file.

lebigot:
Thank you for updating the Travis CI configuration. I will activate Travis back.

It would be useful to also do the tests on Windows as you propose. If you can do a pull request, that's great. I like comments in configuration files too, as they help understand the less obvious parts and update the configuration later if needed. :)

Issue #56, opened 2016-09-04
Plans for native Python 3 suport and dropping support for older Python versions


rth:
Currently `Uncertainties` supports python 2.4-2.7 with support of python 3.2-3.3 through `2to3` converter. There is also the Python 2.3 branch that appears to be still periodically synced.
![untitled](https://cloud.githubusercontent.com/assets/630936/18231140/a5d823c6-72b0-11e6-9c4c-dd4d564c1a20.png)

The support of such outdated python versions (e.g. python 2.3 was installed in Ubuntu Hardy LTS released in 2008 which reached end of life in 2013) is fairly unconventional in the scientific python community were currently most packages support 2.6-2.7 and natively 3.2-3.5 (i.e. without using 2to3), cf. for instance the [last numpy release](https://docs.scipy.org/doc/numpy/release.html) which is also an optional dependency. Also see  the [Python 3 statement](https://python3statement.github.io/) for the dropping of support for PY2 in major scientific python packages by 2020. 

The ongoing support for outdated python versions 2.4-2.5, 2.3 and to a lesser extent 2.6, and the lack of a native 3.2+ support, prevents from using the last features and optimizations of the language  (also related to issue #52) and  in my opinion might also dissuade new contributors who might have already switched to PY3. 

Is there any plans to add native Python 3.2-3.5 support and drop support for older Python versions in the future releases? I would be happy to work on this.

lebigot:
Interesting question.

Thank you for suggesting to work on the native 3.2–3.5 support.

While I do appreciate Python 3 compared to Python 2, my plan is to progressively move to higher and higher versions of Python when I feel that the code becomes too suboptimal or less legible. This is what happened a month ago when I switched the main version to Python 2.7 (instead of Python 2.6)—with older versions supported through an ad hoc branch (master_python2.3), and more recent versions supported with 2to3.

I would guess that it is better if I do the conversion myself, as I know the code well enough to have many of the ideas that could make it more adapted to Python 3. If the main version becomes Python 3 through external help, I am not sure that I will be able to easily track obvious areas where the code could be updated (because 2to3 highlights these areas for me, and it could not be used as easily anymore).

Also, in the past, I have often encountered obsolete Python installations in computing clusters or isolated machines used by larger scientific collaborations. Staying with Python 2 as the native version makes adapting the code to these older version easier—thereby increasing the chances that I don't shut some people out.

[...]

Issue #52, opened 2016-08-18
Make the whole code of the Python 2.7 version use Python 2.7's features


lebigot:
The Python 2.7 code is mostly based on the last Python 2.6 and even Python 2.5 version. Some details of the code could be updated so as to take advantage of the features that Python 2.7's brought, or so as to be adapted to Python 2.7 directly (parts of the code handle differences between Python 2.6 and Python 2.7, for instance).

Examples:
- Dictionary comprehension,
- Sets literals,
- String formatting with `{}`.

The code tends to contain comments like `Python 2.5` or `Python 2.6` or `Python 2.7+` in places where the code could be upgraded.

lebigot:
Was this already done when the code base moved to Python 3?

PR #62, opened 2016-09-16
Updating Travis CI setup

rth:
This partially addresses issue #61 ,

-  fixes and updates the `.travis.yml`: The [tests pass](https://travis-ci.org/rth/uncertainties) on Python 2.7, 3.3, 3.4, 3.5 and for different numpy versions. For 2.6 tests fail with the current code base due to the use of dictionary comprehension in core.py (line 947), not sure if you want to support 2.6 (I would have dropped it as discussed in issue #56, also relevant to issue #52 ). The 3.2 is not tested (as running an exhaustive testing of all versions might take too long), but it could be added.

- Un-comments the Travis-CI badge in the Readme

Please let me know your feedback on this..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ MANIFEST
 uncertainties.egg-info/
 # For PyCharm (contains project files):
 .idea/
+.cache
+.*.swp
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,28 @@
 # Config file for automatic testing at travis-ci.org
-
 language: python
-
-python:
-    # Travis does not have Python 2.4 and 2.5:
-    # - 2.4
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@ MANIFEST
 uncertainties.egg-info/
 # For PyCharm (contains project files):
 .idea/
+.cache
```
lebigot: What's that for (as well as the next line)? I would like to document it.

rth:
Thanks for the feedback. Aww, I was just going to add a comment in the `.gitignore` to say what those extra lines were for, but since it's merged, never-mind, I'll add it in the next PR. Thanks.

PR #63, opened 2016-09-16
Adds CI on Windows 

### LINKED ISSUE #61: Continious Integration

rth:
Fixes #61.

The Appveyor setup is adapted from [here](https://github.com/bsmurphy/PyKrige/blob/master/appveyor.yml) not sure were that one came from, but it is fairly close to examples that could be found [elsewhere](http://tjelvarolsson.com/blog/how-to-continuously-test-your-python-code-on-windows-using-appveyor/), although it is not that readable. It also comes with a compiler in case we need it in the future.

The `lib1to2` fail on Python 3 so I have commented them out (same as in travis) which means that that feature is not tested at all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,7 @@ MANIFEST
 uncertainties.egg-info/
 # For PyCharm (contains project files):
 .idea/
+# py.test cache files (normally we are using nose though)
 .cache
+# vim temporary files
 .*.swp
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,15 +11,17 @@ environment:
     - PYTHON_VERSION: 2.7
       PYTHON_ARCH: "64"
       MINICONDA: C:\Miniconda-x64
-    - PYTHON_VERSION: 2.7
-      PYTHON_ARCH: "32"
-      MINICONDA: C:\Miniconda
-    - PYTHON_VERSION: 3.5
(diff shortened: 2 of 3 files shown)


lebigot:
Do you know by any chance how to run the tests with Python 2.6? Simply setting the version in `appveyor.yml` fails with a message about Python 2.6 being incompatible with NumPy 1.10 (which should actually work, as far as I know). There are also some strange (?) appearances of Python 2.7 in the build process…

rth:
Yeah, just changing the python version in `appveyor.yml` should work, it's just that numpy is installed through conda (i.e. prebuild binary versions) and the highest numpy version build for PY 2.6 is numpy 1.9.2 . So dowgrading to that version should work I think.

lebigot:
This works, thanks. The test now run. But one fails: https://ci.appveyor.com/project/lebigot/uncertainties/build/1.0.14#L284

On OS X with Python 2.6, I get the expected result:
```
>>> "{0}".format(uncertainties.ufloat(-float("inf"), 100))
>>> '-inf+/-100.0'
```

Issue #70, opened 2017-08-15
Support for groups of errors (e.g. statistical and systematic)


nasfarley88:
I propose something like the following:

```python
import uncertainties as uc
x = uc.ufloat(1, 0.5, "x_error_name", error_group="stats")
y = uc.ufloat(-1, 0.7, "y_error_name2", error_group="syst")
z = x + y

print("z: {:L}".format(z))
print("z.std_dev: {}".format(z.std_dev)) # Acts as the independent sum in quadrature of all error groups
```
would print something like
```
z: 0 \pm 0.5 \pm 0.7
z.std_dev: 0.86023252670426
```

This would provide a few advantages over manually calculating separate errors including having proper error propagation for both groups independently.

Changes/improvements to syntax/methods names/etc. are welcome.

lebigot:
Thanks, that's an interesting use case.

The interaction of a form like 0±0.5±0.7 with formatting is not fully trivial I guess (because the nominal value and the uncertainty use the same format, so here we would have to find a meaningful common format for three numbers), but it is worth investigating.

Now, what you propose should definitely be complemented by a way of specifying the _order_ in which each group prints their uncertainty (or we don't know which error comes from where, which makes the feature less useful).

I would suggest to proceed as follows:

1. Maybe a first attempt at providing this kind of feature would be to not go through a new format, but to provide some convenient access to each group error and let the user print them. That should be quite easy to implement.

2. A next step could be to offer some formatting facility so that the nominal value and errors are printed in some particularly legible way (maybe like 1.234 ± 0.021 ± 0.001?).

3. Then later a formatting option could be considered.

I am not promising anything in terms of timeline (I want to implement some speed up idea that I worked on more recently  #(53), and I feel that providing some deeper support for arrays is more urgent). However, adding the concept of error groups and providing some interface to them should be quite doable for external contributors: pull requests are welcome!

PR #72, opened 2017-08-24
Added method error_contribution_from.

nasfarley88:
The function error_contribution_from allows developers to discover the complete contribution from a set of tags specified by a predicate function.

This is part way to address #70. Since Python is awesome (or uncertainties' authors are, reader's choice), tuples can be used for tags instead of strings. This means that with the new function `error_contribution_from` one can:

```Python
from uncertainties import *
x = ufloat(1, 1, tag=("x", "statistical"))
y = ufloat(1, 1, tag=("y", "statistical"))
z = ufloat(1, 1, tag=("z", "systematic"))

w = x + y + z
w_systematic_error = w.error_contribution_from(lambda t: "systematic" in t)
w_statistical_error = w.error_contribution_from(lambda t: "statistical" in t)
```

# Note
This pull request does not (yet!) include documentation changes. These will be added when the code is agreed upon.


nasfarley88:
@lebigot Any thoughts?

lebigot:
Sorry for the delay in responding.

This function can really be a simple one-liner, so I prefer not to include it:

```
def error_contribution_from(self, filter_tag):
    """
    Return the error contribution from specific error components.

    Return the error contribution from error components for
    which filter_tag(tag) is true.
    """
    return sqrt(sum(error**2 for (var, error) in self.error_components().iteritems() 
                    if filter_tag(var.tag)))
```

nasfarley88:
I can see where you're coming from (keeping libraries small is good) but I think a function like this makes uncertainties _much_ more valuable. Relying on the user to figure this out and reimplement it every time instead of providing it in the interface is an unnecessary barrier to entry.

I'm not wedded to the idea of using a function like this; I think there must be a cleaner/clearer way to implement feature. Either way, I think it's worth thinking about how to expose this functionality straight from a `ufloat`. (Maybe through multiple errors per value?)

I use this feature every day for my analysis and it's extremely useful (saving a lot of time).

nasfarley88:
Perhaps an `extras` library would be good for stuff like this? Then there can be these helpful functions without clogging up the main library.

lebigot:
I like the idea of extending the UFloat class with this, thanks! I'm reopening this pull request as a kind of reminder.

PR #79, opened 2018-03-06
Fixes massive FORK ME overlay

naught101:
The FORK ME image is about 50% of the width of the page on firefox on linux at 1080p. This fixes it.

Example screenshot: 
screenshot_20180307_095440

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -6,7 +6,7 @@
 
 {% block document %}
 {{ super() }}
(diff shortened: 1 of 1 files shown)


lebigot:
Thanks.

After your change, is the top-right "index" link clickable? It was not accessible anymore, in Safari on OS X, before my last update of the "Fork me" banner.

Also, which version of Firefox are you using?

lebigot:
Can you check if the changes I propose at https://github.com/lebigot/uncertainties/pull/81 make the "Fork me" banner correctly display on your machine?

I basically added two explicit sizes for the "uncertainties" logo and the banner.

lebigot:
Fixed in current version.

PR #81, opened 2018-04-08
Attempt at controlling the height of the "Fork me" banner better

lebigot:
This hopefully helps some Firefox versions to display the "Fork me" banner with a reasonable size.


lebigot:
This was fixed in the same way in commit 972d67ccf4af10b2.

PR #88, opened 2018-10-11
Use getfullargspec, fix #60

### LINKED ISSUE #60: uncertainties.core - inspect.getargspec() is deprecated
There is a call to `inspect.getargspec()` on line 537 of uncertainties.core.py. Apparently this is deprecated and inspect.signature() should be used instead. Python version 3.5.2

jccurtis:
Some background: We are using this package extensively in a [gamma-ray spectroscopy package](https://github.com/lbl-anp/becquerel) and run into these warnings a lot during testing on `py36`.

As noted in #60, `inspect.getargspec` is deprecated in `py35` and onward. This short PR patches this warning for `py27` and `py36` onward (see [this doc](https://docs.python.org/3.6/library/inspect.html#inspect.getfullargspec) regarding reverting the `DeprecationWarning` between `py35` and `py36`). Additionally, the `inspect.signature` module was added in `py33`.

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -31,6 +31,11 @@ import inspect
 import numbers
 import collections
 
+if sys.version_info < (3,):
+    from inspect import getargspec
+else:
+    from inspect import getfullargspec as getargspec
+
 # Attributes that are always exported (some other attributes are
 # exported only if the NumPy module is available...):
 __all__ = [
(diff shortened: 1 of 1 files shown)


jccurtis:
Apologies, I didn't sufficiently test on `py27`. I have applied a patch. Travis doesn't not want to test with `py33` which appears to be an unrelated issue.

lebigot:
The warning should be removed indeed. Thanks for the pull request. I'll have a look soon.

cdeil:
I'm also getting this warning. Suppressing it for now. Would be nice if this PR got in.

jccurtis:
Awesome, thanks @lebigot!

lebigot:
Thank _you_!

The fix is in version 3.0.3. I will update shortly the package on PyPI.

Now, strangely enough, I can't observe the warning with the previous version: how did you observe it (with just an `import uncertainties` or something else)?

Issue #99, opened 2019-06-23
Use Cholesky decomposition for correlated_values


ces42:
Trying to do by hand exactly what `correlated_values(nom_values, covariance_mat)` does I thought of a different approach than the one used in the current implementation:
```
n = len(nom_values)
L = cholesky(covariance_mat)
return nom_values + dot(L, uarray(zeros(n), ones(n))
```
This essentially does the same as `correlated_values` (except for the tags and it returns a array instead of a tuple). It should be faster than diagonalizing the covariance matrix and some quick tests suggest that it might be numerically more precise:  
For `nom_values = ones(10)` and `covariance_mat = scipy.linalg.hilbert(10)` the reconstructed covariance matrix using `covariance_matrix` only differs in 15 using cholesky while the current implementation differs in 78 entries.

Is there a reason why the current implementation uses diagonalization instead of cholesky? (If there is no such reason I would be happy to write a pull request)

P.S. I really like this package and it's saving me a ton of time in my lab course :)

lebigot:
That's an interesting point: thanks for sharing!

I feel that what you propose is the better way, indeed. A pull request would be much appreciated. 

I am glad that you find this library useful!

PR #103, opened 2019-08-01
Use cholesky decomposition in correlated_values and correlated_values_norm

ces42:
This uses a Cholesky decomposition instead of diagonalization in `correlated_values` and `correlated_values_norm`. The idea is the same as presented in #99.

However, dealing with degenerate covariance matrices proved to be a bit tricky, since numpy refuses to do a Cholesky decomposition of a matrix that is only positive *semi*-definite. Therefore I added a function that performs a LDL decomposition in those cases (this would also be available in scipy).

Local tests (and theory) suggest that this should be more precise and faster than the old implementation.

Both `correlated_values` and `correlated_values_norm` now raise an error if the covariance/correlation matrix fails to be (nearly) positive semi-definite.

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -195,7 +195,7 @@ else:
 
             a = A[i, i]
             l = A[i+1:, i]
-            if a < -EPS or (a <= 0 and abs(l).max() >= EPS):
+            if a < -EPS or (a <= 0 and len(l) > 0 and abs(l).max() >= EPS):
                 raise numpy.linalg.LinAlgError('matrix must be positive '
                     'semidefinite (failed on %s-th diagonal entry)' % i)
 
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -157,7 +157,7 @@ else:
 
         # We perform a cholesky decomposition of the covariance matrix.
         # If the matrix is only positive semidefinite numpy will refuse to
-        # perform a cholesky decomposition so we 'manually' do a LDL
+        # perform a cholesky decomposition, so we 'manually' do a LDL
         # decomposition
         try:
             L = numpy.linalg.cholesky(covariance_mat)
@@ -165,9 +165,11 @@ else:
             L0, D = ldl(covariance_mat)
             L = L0 * numpy.sqrt(D)
 
+        # Creation of new, independent variables:
         if tags is None:
             tags = (None, ) * len(nom_values)
-        variables = [Variable(0, 1, tag) for tag in tags]
+        variables = tuple(Variable(0, 1, tag) for tag in tags)
+
         return nom_values + numpy.dot(L, variables)
 
     __all__.append('correlated_values')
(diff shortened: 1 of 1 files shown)


jagerber48:
I want to make this change, but I think this PR is too old to do so. I think it would be challenging to merge it in. I'll explore a new PR that lifts the main changes from this PR. @ces42 if you're still interested in this you should feel free to open a new PR also.

newville:
@jagerber48 @ces42  I'm +1 on trying to resurrect this.  
I might say both that a) it is not the highest priority, and b) it should not have to wait another 5 years ;).

ces42:
I could try to make an analogous PR again. The main issue still is that numpy's Cholesky decomposition does not work in the semi-definite case, so some kind of fallback is necessary (and creating the fallback will probably be more work than the "general case"). I can see 3 options for the fallback:
1. Use the LDL decomposition from scipy. This would add scipy as a dependency.
2. Use diagonalization as in the current version of `uncertainties`. This is slow and numerically imprecise.
3. Hand-code a LDL (or similar) decomposition, as I ended up doing in this PR. This is slow and adds a lot of code.

It would be possible to detect whether scipy has imported and only do 1. in that case.
Any thoughts on what approach might be best?

jagerber48:
> I might say both that a) it is not the highest priority, and b) it should not have to wait another 5 years ;).

@newville 
Yes, the reason these blocks of code are popping up is because I think they're going to need to be worked on for the large `AffineScalarFunc` refactor but I don't want to either

- rewrite them using their same logic or
- also load the transition to Cholesky decomposition into that PR

so I'm exploring knocking the Cholesky decomposition out first.

________
@ces42 
> Any thoughts on what approach might be best?

> 2. Use diagonalization as in the current version of uncertainties. This is slow and numerically imprecise.

I like this option a lot. The semi-definite case would be something like you have a covariance matrix of random variables, but there is some basis in which one or more of the random variables has zero variance? It's a literal edge/degenerate case that hadn't occurred to me but I'm glad you point it out. Since it is an edge case I think it's no problem if users take a performance hit when they hit that case.

> 1. Use the LDL decomposition from scipy. This would add scipy as a dependency.

We're wringing our hands in some other threads about having `numpy` as an optional or required dependency. I think appetite would be pretty low for adding `scipy`.

> 3. Hand-code a LDL (or similar) decomposition, as I ended up doing in this PR. This is slow and adds a lot of code.

If it is slow what is the advantage of this approach over 2.? Maybe it's faster than 2.but slower than 1.? Seems like we could start with approach 2. then evolve to 3. if there's demand for more performant decompositions on these semi-definite cases.

> It would be possible to detect whether scipy has been imported and only do 1. in that case. This might be considered a bad practice though...

I don't think this is a bad practice. It would be a little bit clunky but the logic for semi-definite matrices could be something like:

- exception if `numpy` is not importable
- eigenvalue decomposition if `numpy` is importable but not `scipy`
- `scipy` Cholesky decomposition if `scipy` is importable

For positive-definite it would be
- exception if `numpy` is not importable
- `numpy` Cholesky decomposition if `numpy` is importable

________

While the above is kind of appealing, I'm personally tempted to say just leave `scipy` out of it as a first pass. For years it's been using the slow `numpy` eigen decomposition and it's seemingly been ok for users. Users can always spin their own function using something else if they have a more performant way to do the decomposition and if that matters for them. We can always add the greedy (i.e. taking advantage of the `scipy` resource if available) performance enhancement down the road.

____
Be aware of this PR that is targeting these same functions and shows some of the discussion around optional dependencies. https://github.com/lmfit/uncertainties/pull/268

jagerber48:
Moving forward on this at #271

PR #107, opened 2019-11-18
modify python2 specific lines to python3 standard

BishopWolf:
Python2 lines create a conflict making them cause errors in python3.
Refer to [this issue](https://github.com/lmfit/lmfit-py/issues/605) for a complete description


lebigot:
Thanks but there is already a mechanism in place for generating a Python 3 version (`2to3`, called upon installation), which is tested through [continuous integration](https://travis-ci.org/lebigot/uncertainties). There is no need for doing this manually.

PR #108, opened 2019-11-20
Distribute package as wheel

benabel:
First thank you for this excellent package, I would like to use it in [pyodide](https://pyodide.readthedocs.io/en/latest/index.html) however it needs to be packaged as a wheel. 

Sources:

- https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives

- https://packaging.python.org/key_projects/#twine

### DIFF
--- a/00_prepare_for_PyPI.sh
+++ b/00_prepare_for_PyPI.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-# This script must be run before packaging (twine upload dist/*).
+# This script prepares the package for PyPI. It must be run
+# before uploading it on PyPI.
 
 # WARNING: this script erases any uncertainties-py23 and uncertainties-py27
 # found in the current directory.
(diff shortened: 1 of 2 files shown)


BishopWolf:
[We have been waiting](https://discourse.slicer.org/t/problems-installing-lmfit-python-package/9210) for this wheel also, thank you very much for this PR.

00_prepare_for_PyPI.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# This script must be run before packaging (python setup.py sdist upload).
+# This script must be run before packaging (twine upload dist/*).
```
lebigot: Well spotted!

This instruction generally fails, though: older versions are typically present in `dist/`, and `twine` PyPI rejects them. I'll update this line.

setup.py
```diff
@@ -385,7 +381,7 @@
     # distutils.core.setup is not like setuptools: it does not have an
     # option for automatically calling 2to3 if Python 3 is used, so the
     # conversion is done here:
-    if sys.version_info >= (3,):
+    if sys.version_info >= (3, ):
```
lebigot: This is not standard: this is better reverted.

lebigot:
Thanks a lot for this pull request!

I do have one question: https://github.com/lebigot/uncertainties/pull/109/files includes `if __name__ == "__main__"` at the end of `setup.py`. Do you know if this is needed?

benabel:
@lebigot I don't think `if __name__ == "__main__"` is required as `setup.py` should always be run from python.

lebigot:
Thanks. I'll leave it this way then.

Issue #114, opened 2019-11-30
numpy Deprecation warning


benabel:
`uncertainties.unumpy.array` throws `DeprecationWarnings` from `numpy`:

    PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    mat = unumpy.matrix(arr)

jonasstein:
please add the exact version information for numpy and uncertainties to your initial ticket (edit).

lebigot:
Thanks for reporting this.

Handling this requires some care: [users expect](https://uncertainties-python-package.readthedocs.io/en/latest/numpy_guide.html#matrices) `unumpy.matrix` objects to behave like NumPy matrices.

I'm thinking that uncertainties should deprecate `unumpy.matrix` in the same way as NumPy is doing.

Another thing that should be done is to see how to replace the `.I` matrix inversion.

benabel:
So you'd create a new type unumpy.ndarray that uses numpy arrays and deprecate unumpy.matrix with warnings. 

I think inversion can bé done with numpy linalg module https://numpy.org/devdocs/reference/generated/numpy.linalg.inv.html#numpy.linalg.inv

lebigot:
Yes, indeed.

`uncertainties` does provide the more powerful `pinv()` already.

AdityaSavara:
Hello, I found the same error and I am using pinv, but I have also been using umatrices for a dot product. I have not found any linalg function or unumpy function to do a dot product with the uarray objects.  Is there a built in function to do dot products with the uarray objects? If not, maybe a volunteer should offer to help build this in for when umatrix becomes deprecated.

lebigot:
Have you tried using usual numpy.dot() or the @ operator as usual? There is no "uarray" class: the uarray() factory returns a regular NumPy array, which should behave as expected (in many contexts).

AdityaSavara:
I'm sorry for my delay in responding, I needed to wait until I got home from work to check due to which computer the code is on.  I had tried syntax like "unumpy.dot()" but had not tried syntax like "numpy.dot()".  I am surprised to see that even using the @ operator with uarrays does work. It might be good to provide an @ operator example in the documentation (or to mention it).

lebigot:
[...]

PR #116, opened 2019-12-06
Add enclosing round bracket flags r and R to the format_num method

ep12:
Implement a new formatting option, see #113.
Note: This doesn't work with unumpy arrays like regular formatting does not work with them, formatting in general does not work with them (regular numpy's fault?).

PR #120, opened 2020-02-11
 Removing unumpy.py core dependency on matrices 

AdityaSavara:
Removing numpy deprecation warning mentioned here #114

Changed "asmatrix" to "asarray"
Changed * operator to @ operator for the inverse functions (as shown in diff file below).
Passed unit tests on my computer except for the lib2to3 unit test since I do not have that module.
This also removed the deprecation warnings during unit tests on my code, without breaking my own code's unit tests.

Adding .conj() for non-real numbers. 

I don't have a unit test to check correct the behavior for non-real numbers.

### DIFF
--- a/uncertainties/unumpy/core.py
+++ b/uncertainties/unumpy/core.py
@@ -462,12 +462,12 @@ def inv_with_derivatives(arr, input_type, derivatives):
     yield inverse
 
     # It is mathematically convenient to work with matrices:
-    inverse_mat = numpy.asmatrix(inverse)
+    inverse_mat = numpy.asarray(inverse)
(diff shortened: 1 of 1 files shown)


AdityaSavara:
I am not sure if the integration checks are failing because I used the @ operator or due to something else. Would it be better to  numpy.matmul?

ep12:
> Would it be better to use numpy.matmul?  

Probably not, the errors in the logs seem to not be related to your changes. I guess those errors affect not only this pull request. I think my pr (#116) fixes some of the bugs because they annoyed me a lot...

Issue #121, opened 2020-02-25
norm of vector with uncertainties


karlwessel:
Calculating the norm of a numpy array with uncertainties using numpy.linalg.norm throws an error which can be reproduced with the following MWE:
```
from uncertainties import unumpy
import numpy
a = unumpy.uarray([1, 1], [1, 1])
numpy.linalg.norm(unumpy.nominal_values(a)) # works fine
numpy.linalg.norm(a)
```
This throws the following exception:
```
AttributeError: 'AffineScalarFunc' object has no attribute 'sqrt'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 6, in norm
  File "lib/python3.6/site-packages/numpy/linalg/linalg.py", line 2483, in norm
    ret = sqrt(sqnorm)
TypeError: loop of ufunc does not support argument 0 of type AffineScalarFunc which has no callable sqrt method
```

This is with uncertainties version 3.1.2 and numpy version 1.17.4.

lebigot:
Thanks for reporting this. Either we should have a unumpy.linalg.norm function or, better, some mechanism that makes the original numpy.linalg.norm work.

karlwessel:
I added a pull request that fixes the problem. However, since it adds sqrt as a member function to AffineScalarFunc it might introduce some unexpected side effects.

lebigot:
Thanks. I am quite confident that this should work fine. I'll have to review the pull request when I am close to my personal computer but so far things look good.

lebigot:
Thanks. I am quite confident that this should work fine. I'll have to review the pull request when I am close to my personal computer but so far things look good.

PR #122, opened 2020-02-25
Add sqrt member to AffineScalarFunc to make numpy.linalg.norm work

karlwessel:
This is my own try at fixing issue #121. 

It works by adding `sqrt` as a member function to `AffineScalarFunc` so that `numpy` knows how to calculate the square root of it.

This also makes some other `numpy` methods work like `numpy.sqrt`.

### DIFF
--- a/uncertainties/test_umath.py
+++ b/uncertainties/test_umath.py
@@ -341,3 +341,21 @@ def test_power_wrt_ref():
     Checks special cases of the umath_core.pow() power operator.
     '''
     test_uncertainties.power_wrt_ref(umath_core.pow, math.pow)
+    
+def test_numpy_sqrt():
+    '''
+    Checks if numpys sqrt method works due to AffineScalarFunc having basic math
+    functions as members.
+    '''
+    try:
+        import numpy
+    except ImportError:
+        import warnings
+        warnings.warn("Test not performed because NumPy is not available")
+        return
+        
+    a = ufloat(0.3, 0.01)
+    if not(numpy.sqrt(a) == umath_core.sqrt(a)):
+        raise Exception('Numpy sqrt (%s) and math sqrt (%s) do not match!' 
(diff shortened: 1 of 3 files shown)


pkienzle:
The method used here (adding `sqrt` to the class) will work for many (all?) of the functions in umath, so for example `np.sin(ufloat)` would return `umath.sin(ufloat)`.  Any interest in extending this patch to include all these functions?

The attribute would be needed for both `Variable` and `AffineScalarFunc`.

Some methods will need different names, such as `arcsin` instead of `asin`.

karlwessel:
> The method used here (adding `sqrt` to the class) will work for many (all?) of the functions in umath, so for example `np.sin(ufloat)` would return `umath.sin(ufloat)`. Any interest in extending this patch to include all these functions?
> 
> The attribute would be needed for both `Variable` and `AffineScalarFunc`.
> 
> Some methods will need different names, such as `arcsin` instead of `asin`.

During implementation of the patch I already did that for `exp` but removed to keep the PR on point (which is just fixing issue #121).

I think once this PR is merged and has proven to work for some time, one could implement your suggestion in a separate PR.

AndrewAnnex:
I can see that some time has past since this pr was last updated, and I just ran into this issue (plus the norm one) when attempting to use uncertainties with statsmodels. Are the failed checks related to an incomplete implementation? If this works as-is I can make progress on my own work by just monkey-patching

AndrewAnnex:
I also see an issue where numpy needs an implementation of the conjugate method as the default in numpy is to use the  Frobenius norm, although this is really not my area

jagerber48:
My rewrite, discussed in #251 implements `sqrt` on `UFloat` and so would cover this PR and associated issue. In that rewrite the `np.linalg.norm` functions work on a `dtype=object` `ndarray` of `UFloat` objects, or on the `UArray` object (thin wrapper for `dtype=object` `ndarray` of `UFloat`).

Issue #113, opened 2019-11-27
bracket flag


ep12:
I wonder if it would be possible to add a `b` (bracket) flag as a format specifier to always have enclosing brackets (force!) with a common factor (optionally). This would be heavily useful for stuff with units:
The correct format is `(0.8265±0.0735)×10⁻³ m = (0.8265±0.0735) mm`, not `0.8265±0.0735 mm`.
The right place to check for the common factor / enclosing brackets is this package I think, not in the user's module.

lebigot:
Do you have a reference about the correct format having parentheses? (This vaguely reminds me of a discussion one of my colleagues at the National Institute of Standards and Technology had, one day, arguing that the part before "±" should have explicit units!)

I do agree that if you want to ensure that the printed value does have parenthesis, this should be handle by the uncertainties package as a way of formatting numbers with uncertainty.

ep12:
No, I don't have a written reference about the correct format, but it is obvious that parens are needed when having non-1 units. It simply does not make sense if the uncertainty has a different unit (e. g. 1) than the n value. That is equivalent to the addition (subtraction) of two values with different units. What is a second added to a meter? What the sum of 5 lux and 6 degrees?
If we have `x = 5 m +/- 0.2 m` everything works and we can factor: `x = (5 +/- 0.2) m`. `x = 5 m +/- 2` might be intuitively interpreted right, but it is formally wrong. Another similar example: What is the sum of a vector `x ∈ ℝ⁴` and another vector `x ∈ ℝ⁵`? The sum `+: ℝ⁴×ℝ⁵ → ??` simply is not defined! You can think of units as something quite similar to the imaginary unit `𝚤` (just the _unit_ properties, without the special properties, e.g. `𝚤²=-1`): `5 + 3𝚤` is just `5 + 3𝚤`. I hope you get my point why the parentheses are absolutely required for formal reasons...?

ep12:
The pull request (#116) I created  is ready for review, I think.
- [x] add `b` and `B` format flags
- [x] add tests
- [x] add a simple check for surrounding parenthesis to `ufloat_fromstr`
- [x] minor changes (mostly unicode related, one TypeErrors, a python2 error) to make the builds pass
I know it's not good to have those CI test fixes in a pr with a different topic, however the only commits needed to implement the flags are 053ae8e and 1600fa7.

lebigot:
Thanks for the pull request.

[...]

PR #128, opened 2020-08-02
R flags

ep12:
See #113.

### DIFF
--- a/doc/user_guide.rst
+++ b/doc/user_guide.rst
@@ -493,6 +493,22 @@ formatting options. They can be added at the end of the format string:
   >>> print '{:L}'.format(x*1e7)  # Automatic exponent form, LaTeX
   \left(2.00 \pm 0.10\right) \times 10^{6}
 
+- ``b`` for surrounding **brackets** (common exponent outside)
+
+	>>> print '{:b}'.format(x)
+	(2.00+/-0.10)e-01
+- ``B`` for surrounding **brackets** (common exponent inside)
+
+	>>> print '{:B}'.format(x)
+	((2.0+/-0.10)e-01)
+The ``b`` and ``B`` options can be useful when dealing with units:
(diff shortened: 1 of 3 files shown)


lebigot:
I would have a few fixes and improvements to make. Apparently, you could allow me to push to this specific pull request branch: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork. Can you enable this?
Otherwise, I can always import locally your r-flags branch, update it and then merge it back to master, but this pull request would have to be closed without appearing to ever have been merged.

ep12:
Just in case you missed it: I enabled the editing so that you can change the code to your liking.

lebigot:
I have been working on this here and there. Here are some highlights of my changes so far:
- Switch from an r/R formatting flag to a p (parentheses) flag.
- Fixed inconsistencies between b/B and r/R flags (both were used).
- Fixed bug with print('{:.1u%p}'.format(x)): spurious second parenthesis.
- Better doc.

Working branch: https://github.com/lebigot/uncertainties/tree/ep12_r-flags (I saw your message about the access rights on your fork only after starting working on a local branch).

There are still parts of the code that I need to proofread (notably the tests).

lebigot:
I kept checking the changes and updated a few more things:
- Fixed bug with "  (2+/-1)  " not being parsed by `ufloat_fromstr()` (contrary to what the documentation claimed).
- Added a test for the bug above.
- Moved some new code and doc where they belong (parsing done inside the existing parsing function instead of outside, detailed documentation for the formats in the more user-facing `AffineScalarFunc.__format__()`, etc.).
- Removed duplicated documentation for the new "p" formatting option (it's still duplicated between the detailed web user guide and the docstrings, but that's expected).

PR #138, opened 2021-06-26
make changing the pretty-print times character easy

ari-s:
changing the × in pretty printing to me seems like it should be easy.
It's not hard now, but I had to get into the code. With this PR, it becomes:
```python
uncertainties.core.PRETTY_PRINT_TIMES = "⋅"
```

If you would take this PR, I guess the documentation should also mention it: currently it mentions that it's possible to change pretty printing, but not how. If not mentioned, it would be easy to do, but not easy to discover.

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -1021,10 +1021,14 @@ def nrmlze_superscript(number_str):
     # a subclass of unicode, in Python 2):
     return int(str(number_str).translate(FROM_SUPERSCRIPT))
 
+# 'times'-symbol used for pretty-printing in the EXP_PRINT function below
+PRETTY_PRINT_TIMES = u'×'
+
 # Function that transforms an exponent produced by format_num() into
 # the corresponding string notation (for non-default modes):
 EXP_PRINT = {
-    'pretty-print': lambda common_exp: u'×10%s' % to_superscript(common_exp),
+    'pretty-print': lambda common_exp: u'%s10%s' % (
+        (PRETTY_PRINT_TIMES, to_superscript(common_exp))),
     'latex': lambda common_exp: r' \times 10^{%d}' % common_exp}
 
 # Symbols used for grouping (typically between parentheses) in format_num():


lebigot:
This is a reasonable need, so I'm all for giving this kind of flexibility.

Now, I wouldn't want LaTeX users to not have the same flexibility, so it would be best to also have the LaTeX option listed, with something like `PRINT_TIMES = {'pretty-print':…, 'latex':…}`, and the corresponding setting `uncertainties.core.PRINT_TIMES['pretty-print'] = "⋅"`.

I can do this.

And, yes, indeed, this could also be documented. At least the doc shouldn't say it can be configured if it cannot be… I'll have a look.

lebigot:
Where does the documentation say that pretty printing can be customized? I don't remember putting this there, and I can't find it…

ari-s:
> Where does the documentation say that pretty printing can be customized? I don't remember putting this there, and I can't find it…

It's somewhat hidden.
```python
help(uncertainties.UFloat.__format__)
'''
[...]
    Some details of the formatting can be customized as described
    in format_num().
'''
help(uncertainties.core.format_num)
'''
[...]
The formatting can be partially customized globally.  The EXP_PRINT
maps non-default modes ("latex", "pretty-print") to a function
that transforms a common exponent into a string
[...]
'''
```

Also mentioning this in the user guide would be nice as well.

### Smalltalk

> This is a reasonable need, so I'm all for giving this kind of flexibility.

Great to hear!

> Now, I wouldn't want LaTeX users to not have the same flexibility, 

I'm not in academia anymore, so I don't need to produce printable stuff, so no personal need for Latex anymore :) Not doing latex is something I really enjoy ;) And yet I get to do some statistics now and then.

lebigot:
I included this pull request, thanks.

I added quite many things to it (customization of the ± symbol as well, LaTeX customization, and documentation for all of this).

This is in https://github.com/lebigot/uncertainties/releases/tag/3.1.6 (also on PyPI).

ari-s:
thank you for your great work in general!

PR #146, opened 2021-10-27
ENH: only import from past on Python 2

### LINKED ISSUE #145: dependency on past generates warnings in Python 3
`uncertainties` imports `imp` via `past` at import time, but `imp` is deprecated from the standard library, which means that in the current state, uncertainties will not be compatible with Python 3.12
This is a problem for me as I want to future-proof a package I'm maintaining which depends on `uncertainties`, so I'd like to treat deprecation warnings as errors, and don't want to ignore the ones coming from uncertainties.
I believe the patch is trivial so I'll give it a go but I wanted to open an issue first for reference.

reproduction:

```python
# t.py
import uncertainties
```

and running on the command line
```bash
python3 -Werror t.py
```

yields
```
Traceback (most recent call last):
  File "/Users/robcleme/dev/uncertainties/t.py", line 1, in <module>
    import uncertainties
  File "/Users/robcleme/dev/uncertainties/uncertainties/__init__.py", line 225, in <module>
    from .core import *
  File "/Users/robcleme/dev/uncertainties/uncertainties/core.py", line 22, in <module>
    from past.builtins import basestring
  File "/Users/robcleme/.pyenv/versions/uncertainties-dev/lib/python3.9/site-packages/past/builtins/__init__.py", line 54, in <module>
    from past.builtins.misc import (apply, chr, cmp, execfile, intern, oct,
  File "/Users/robcleme/.pyenv/versions/uncertainties-dev/lib/python3.9/site-packages/past/builtins/misc.py", line 45, in <module>
    from imp import reload
  File "/Users/robcleme/.pyenv/versions/3.9.7/lib/python3.9/imp.py", line 31, in <module>
    warnings.warn("the imp module is deprecated in favour of importlib; "
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
```

neutrinoceros:
Fix #145

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -19,7 +19,6 @@ from builtins import next
 from builtins import map
 from builtins import zip
 from builtins import range
-from past.builtins import basestring
 from builtins import object
 import math
 from math import sqrt, log, isnan, isinf  # Optimization: no attribute look-up
@@ -45,6 +44,14 @@ if sys.version_info < (3,):  # !! Could be removed when moving to Python 3 only
     from inspect import getargspec
 else:
     from inspect import getfullargspec as getargspec
+
+if sys.version_info < (3,):
+    from past.builtins import basestring
+else:
+    # avoid importing from past in Python 3 since it utilizes the builtin 'imp'
+    # module, which is deprecated as of Python 3.4, see
+    # https://docs.python.org/3/library/imp.html
+    basestring = (str, bytes)
 
 # Attributes that are always exported (some other attributes are
 # exported only if the NumPy module is available...):


neutrinoceros:
I'll close this to clear off my personal backlog. Feel free to reuse my this if you want.

lebigot:
Thank you.

I understand that for Python 3, `basestring = (str, bytes)` should instead be `basestring = str`. In fact, this is [what 2to3 does](https://docs.python.org/3/library/2to3.html?highlight=basestring#to3fixer-basestring). This should more importantly work in the code because this should be the type of a keyword argument (key).

I'll include your patch and change this. Thanks again!

lebigot (commented):
Thank you.

I understand that for Python 3, `basestring = (str, bytes)` should instead be `basestring = str`. In fact, this is [what 2to3 does](https://docs.python.org/3/library/2to3.html?highlight=basestring#to3fixer-basestring). This should more importantly work in the code because this should be the type of a keyword argument (key), and can therefore be simply `str`.

I'll include your patch and change this. Thanks again!

PR #170, opened 2023-07-05
Make AffineScalarFunc hashable for Pandas, Pint and Pint-Pandas

MichaelTiemannOSC:
Add __hash__ to AffineScalarFunc so it works with Pandas, Pint, and Pint-Pandas.

Part of https://github.com/pandas-dev/pandas/pull/53970
and https://github.com/hgrecco/pint/pull/1615
and https://github.com/hgrecco/pint-pandas/pull/140

 @jbrockmendel @hgrecco @andrewgsavage

### DIFF
--- a/uncertainties/test_uncertainties.py
+++ b/uncertainties/test_uncertainties.py
@@ -2404,3 +2404,21 @@ else:
         assert arrays_close(
             numpy.array(cov_mat),
             numpy.array(uncert_core.covariance_matrix([x2, y2, z2])))
+
+    def test_hash():
+        '''
+        Tests the invariance that if x==y, then hash(x)==hash(y)
+        '''
+
+        x = ufloat(1.23, 2.34)
+        y = ufloat(1.23, 2.34)
+        # nominal values and std_dev terms are equal, but...
+        assert x.n==y.n and x.s==y.s
+        # ...x and y are independent variables, therefore not equal as uncertain numbers
+        assert x != y
+        assert hash(x) != hash(y)
+
+        # the equation (2x+x)/3 is equal to the variable x, so...
+        assert ((2*x+x)/3)==x
+        # ...hash of the equation and the variable should be equal
+        assert hash((2*x+x)/3)==hash(x)
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -19,10 +19,4 @@ or, if additional access rights are needed (Unix):
 testing framework.  This can be achieved for instance with a command
 like
 
-  nosetests -sv uncertainties/
-
-or simply
-
-  nosetests uncertainties/
-
-(for a less verbose output).
+  pytest uncertainties/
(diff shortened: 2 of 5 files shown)


MichaelTiemannOSC:
While I solved the invariant problem, when I ran pytest the pickle test failed.  All others passed.  I'll see what I can do about that...

uncertainties/test_uncertainties.py
```diff
@@ -2404,3 +2404,21 @@ def test_correlated_values_correlation_mat():
         assert arrays_close(
             numpy.array(cov_mat),
             numpy.array(uncert_core.covariance_matrix([x2, y2, z2])))
+
+    def test_hash():
```
jbrockmendel, line 2408: are there any NaN-likes that will have x!=x?  if you want those to be considered "matching" for e.g. dict lookups (and pandas Index lookups) those will need matching hashes

MichaelTiemannOSC: My understanding is that there is no canonical NaN in the uncertainties package:

```
import numpy as np
import uncertainties

xx = ufloat(np.nan, 0)
yy = ufloat(np.nan, 1)

# nan+/-0
print(xx + xx)
# nan+/-2
print(yy + yy)

# False, because nan != nan
print(xx == xx)
```

The uncertainties package has unumpy.isnan which works for both scalar and arrays.  There's no dictionary compare against a nan.  The PintArray type uses unp.isnan to check whether something is a nan or not.  The pandas changes all us PintArray.isna to test NA-ness (including NaN-ness).  But if I'm missing something, please let me know!

MichaelTiemannOSC: I've now changed the Pint-Pandas changes to use np.nan instead of pd.NA.  While I was able to get everything to work with pd.NA, things were cleaner using np.nan as the NA value for uncertain values.

MichaelTiemannOSC:
All changes needed for Pandas 2.1 are now present and Pint/Pint-Pandas changes are ready for review.  Could we get a fresh release with hashable AffineScalarFunc?

NelDav:
I made similar changes in #189, just because I haven't seeen this MR before. I do not have any clue about uncertainties, but wouldn't it be necessary, to check the correlations as well? You may check my MR, I tried to do it there.

NelDav:
Hi @MichaelTiemannOSC, thank you for your answer. I merged your branch into mine an checked my hash funciton with your tests. Indeed,  a few problems occoured. I then modified my hash calculation and now your tests are pasing. I think the modification is still simpler than your solution. However, I by far not an expert on how do calculate with uncertainties and how the identity check should be beformed. It would be very nice if you could look at my new proposal and share your impressions.

jagerber48:
See my comment on the other PR: https://github.com/lmfit/uncertainties/pull/219#issuecomment-2241430414

Closing this PR in favor of #219. Though that one will likely be closed as well since I think we discovered implementing a hash requires a bit of a larger refactor of the `AffineScalarFunc` and `LinearCombination` code than simply slapping a `__hash__` function somewhere.

PR #174, opened 2023-10-12
Make `str_to_number_with_uncert()` locale-aware.

PeterKraus:
So, `uncertainties` currently cannot parse strings using decimal commas, such as `456,43` into the expected tuple, `(456.43, 0.01)`. This is because the regex on [line 2968](https://github.com/lebigot/uncertainties/blob/804adccf3401aeacbcbae0d669f92131fcd02c03/uncertainties/core.py#L2968) is too restrictive:

```
POSITIVE_DECIMAL_UNSIGNED_OR_NON_FINITE = r'((\d*)(\.\d*)?|nan|NAN|inf|INF)'
                                                    ^--- only allows decimal point, not comma.
```

However, modifying this alone will cause the `float()` conversions to fail. There are two ways around it:
- use `locale.atof()` instead, respecting the currently set locale
- replace all commas for points and keep using `float()`

I suppose doing it the first way (respecting locales) might be the "correct" way of doing it, but it might break things. It's what the current PR does. Please let me know if I should implement it the other way. 

I'll then add tests.

Cheers!

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -18,6 +18,7 @@ from builtins import str, next, map, zip, range, object
 import math
 from math import sqrt, log, isnan, isinf  # Optimization: no attribute look-up
 import re
+import locale as lc
 import sys
 if sys.version_info < (3,):
      from past.builtins import basestring
@@ -159,7 +160,7 @@ else:
         covariance_mat -- full covariance matrix of the returned numbers with
         uncertainties. For example, the first element of this matrix is the
         variance of the first number with uncertainty. This matrix must be a
-        NumPy array-like (list of lists, NumPy array, etc.). 
+        NumPy array-like (list of lists, NumPy array, etc.).
 
         tags -- if 'tags' is not None, it must list the tag of each new
         independent variable.
@@ -167,9 +168,9 @@ else:
 
         # !!! It would in principle be possible to handle 0 variance
         # variables by first selecting the sub-matrix that does not contain
-        # such variables (with the help of numpy.ix_()), and creating 
+        # such variables (with the help of numpy.ix_()), and creating
         # them separately.
-        
+
         std_devs = numpy.sqrt(numpy.diag(covariance_mat))
 
         # For numerical stability reasons, we go through the correlation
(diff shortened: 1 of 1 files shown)


jagerber48:
There's discussion about moving to `sciform` for formatting. For what it's worth, `sciform` has code for parsing inputs into values/uncertainties. From the string itself `sciform` tries to infer whether `.` or `,` is the decimal separator. In cases where it can't do this then it uses either a user-supplied decimal separator or the configured default decimal separator. 

I'm a little wary in `sciform` or `uncertainties` to move from automatically detecing `.` or `,` based on locale largely because I'm a bit unfamiliar with intricacies of the `locale` module but also because I could see surprising things happening.

In the future I see `uncertainties` parsing strings like `123,456.789` and `123.456,789`. That is, strings that can include decimals/commas for both the decimal **and** thousandths separator. Currently uncertainties assumes there is no thousandths separator and that the only decimal separator is a point. This PR allows only a decimal separator but it can be a point or comma.

Here's a few options that don't include thousandths separators:
1. status quo: decimal symbol must be a point, sciform won't accept `1,23`.
2. `uncertainties` always accepts point or comma decimal separators. This would be accomplished by always converting `,` -> `.` before parsing
3. `uncertainties` accepts the locale specific decimal separator.
4. `uncertaintes` accepts only commas as decimal separators
5. `uncertainties` accepts some combinations of the above as selected by user configuration.

My vote would be for option 2. It's straightforward to explain and it would be straightforward to implement. It does get tricky when thousandths separators get included but I think `sciform` shows a path forward where the parsing can get as far as possible on its on and then fallback to a default option when it can't automatically infer it. Perhaps the default options could be configured to be locale specific.

PeterKraus:
Yes, touching `locale` module is tricky, as it's not thread/process safe. I actually swapped all processing in my libs internally to `babel`, which produces a `Decimal` (so you can get the uncertainty from that). However, localised numbers with uncertainties (e.g. `3.565,24(12)`) cannot be processed by `babel`.

I imagine a library that can parse strings into the nominal & error components, which can take a locale specifier (e.g. `de_DE` or `en_US`) to determine the decimal separator, or supply the decimal separator as an option (as `sciform` seems to do) would be quite useful, actually.

I think this PR can be closed either case.

PR #188, opened 2024-01-21
[WIP] Subclassing

andrewgsavage:
This uses the subclassing approach I mentioned in #187, splitting AffineScalarFunc into a few smaller files to make it easier to work with.

For example, if you're looking at formatting you should only need to look at the formatting.py. This is much smaller than the current core.py and only contains code relevant to formatting so you won't get lost looking at other functionality like adding or equality.

The numpy functionality is WIP but IIRC the tests were working without that.


andrewgsavage:
implementing `__array_ufunc__` gets the existing numpy functionality back (unumpy tests pass again). I've just looped over arrays elementwise applying wrapped numpy funcs rather than using the `wrap_array_func` in unumpy.

I think a more generic version of `wrap_array_func` that can accept ndarrays as any arg or kwarg will be needed for the numpy functions that operate on arrays.

PR #196, opened 2024-03-07
Add a job to GitHub Actions aggregating test results

wshanks:
This job succeeds only if all the test jobs succeed. It allows us to mark it as required for merging a PR rather than requiring every individual job.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-
+    name: Tests
     strategy:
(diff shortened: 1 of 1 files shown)


wshanks:
@jagerber48 This is a follow up to #194 just to add an extra check because in the repo settings it seemed like I otherwise had to select every job of the matrix separately as a required status check, which is tedious and brittle (we would probably forget to go into the settings and check the boxes for Python 3.13 when adding that to the matrix).

jagerber48:
Seem fine to me.

PR #198, opened 2024-03-10
Break formatting code into formatting.py

jagerber48:
Move the code for formatting `ufloat` into its own module: `formatting.py`.

I made two tiny moves beyond what I would call pure copy and paste:

1. I always define the `isinfinite` method and never try to import it from `math`.
2. I copy and pasted the `deprecation` function from `core.py` into `formatting.py` to avoid dealing with a circular import. The proper way to do this would be to place `deprecation()` into a common `utilities.py` or similar that both `core.py` and `formatting.py` could import from. But I didn't want to wade into that move in this PR.


newville:
I am not opposed to this change.  I think it can and should wait until after the next release.

andrewgsavage:
Would you like to revive this?

jagerber48:
Yes, I would like to revive this when I have time. I am very busy right now and likely won't have times to get into the details on this for at least a couple of weeks. If anyone else would like to revive it in the meantime instead they should feel free.

Issue #208, opened 2024-03-31
Remove unused imports in tests.


willynilly:
There are two unused imports in the test files:

sys is not used uncertainties/test_umath.py
weakref is not used in uncertainties/test_uncertainties.py

willynilly:
I added a pull request here: https://github.com/lmfit/uncertainties/pull/209

PR #209, opened 2024-03-31
Removed two unused imports in two tests.

willynilly:
I removed the unused imports mentioned in the following issue:

https://github.com/lmfit/uncertainties/issues/208

### DIFF
--- a/uncertainties/test_umath.py
+++ b/uncertainties/test_umath.py
@@ -10,7 +10,6 @@ from __future__ import division
 from __future__ import absolute_import
 
 # Standard modules
-import sys
 import math
 
 # Local modules:
@@ -22,6 +21,7 @@ from . import test_uncertainties
 
 ###############################################################################
 # Unit tests
+
 
 def test_fixed_derivatives_math_funcs():
     """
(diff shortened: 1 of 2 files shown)


andrewgsavage:
this is a lot more than just moving two tests as you've changed line spacing and whitespace throughout those files. I assume you've run this through a linter. We'd like to get a linter running in the CI to give formatting throughout. I think changes to formatting should be done after a linter is set up in CI.

newville:
I agree with @andrewgsavage.  Removing the imports is fine, but that should change 2 lines of code.

willynilly:
i am curious. what is the problem with changing line spaces before you add a linter to the build process? my changes passed all the tests.

jagerber48:
> i am curious. what is the problem with changing line spaces before you add a linter to the build process? my changes passed all the tests.

While your efforts to clean up the code base are much appreciated, we should keep each PR as focused as possible so it's more easier for everyone to follow what changes were made where and when. This PR should handle the imports like its title suggests, but the job of whitespace cleanup should be handled in a dedicated PR. As has been said, there's some likelihood we'll have automated CI in the future which will include formatting. It will be best to bundle all the "formatting" changes into a single PR around that time. 

I understand the temptation to "clean up" things like this as you go. But,, I've come to learn, it causes the repo documentation to get muddled some, and this matters more for a heavily used public repo that a team of people work on. So I think it's best practice to avoid "sneaking" general cleanup into PRs dedicated for something else.

PR #210, opened 2024-03-31
Removed unused imported modules from tests.

willynilly:
Removed unused imports in tests.  

My last pull request was rejected because I had reformatted the file with autopep8 (https://pypi.org/project/autopep8/).  This pull request does not reformat the file with autopep8.

PR #211, opened 2024-04-01
move tests to tests folder

andrewgsavage:


newville (dismissed):
Thanks!! This looks good to me.

andrewgsavage:
> Thanks!! This looks good to me.

This now shows the changes I had made for removing depreciations and 1to2, which I don't think you saw when you commented. I'll open a PR for 1to2 now that can be merged before this.

wshanks:
Did you consider moving the tests out of the package (just a top level `tests/` directory next to `uncertainties/`)? I think that is more standard.

I have occasionally found it useful to have tests in the main package so I can test the package with `pip install package && python -m unittest package.tests` and without separately cloning the repo, but I don't think pytest supports such execution of tests in an installed package. I think it always wants to run from a working directory containing the tests below it.

newville:
+1 on moving tests out of the source tree and into a top-level `tests` folder.  
But then the CI scripts need updating too.

wshanks:
The CI scripts just do a plain `pytest` call, though it would be nice to specify `tests/` so pytest doesn't waste time searching other files (not that there are many here). I mentioned adding a minimal pytest configuration (like [this](https://docs.pytest.org/en/7.4.x/reference/customize.html#pytest-ini)) in https://github.com/lmfit/uncertainties/pull/205#issuecomment-2030023289 because I think it might also help with that case (VS Code knowing to run pytest).

newville:
@wshanks Ah, thanks, yes I see that now, and I completely agree.
I would probably go with the "add pytest config to pyproject.toml", but adding a pytest config file seems okay to me too.

andrewgsavage:
interesting that it keeps the approval after I make new commits

I moved some functions that could be used by downstream libraries to uncertainties.testing

wshanks:
> interesting that it keeps the approval after I make new commits

I enabled dismissing approvals when new changes are pushed. I have found it to be annoying on other projects where the reviewer approves the spirit of the PR but then has to come back and approve a couple more times because of typos/linting. Here, it seems like we have a responsive enough group for it to be okay.

tests/test_umath.py
```diff
@@ -1,25 +1,13 @@
...
 
-# Local modules:
 from uncertainties import ufloat
 import uncertainties.core as uncert_core
 import uncertainties.umath_core as umath_core
 
-from . import test_uncertainties
-
+from uncertainties.testing import compare_derivatives, numbers_close
+from test_uncertainties import power_special_cases, power_all_cases, power_wrt_ref
```
wshanks, line 10: It does not seem right to import from one `test_` file to another. These functions do seem too specialized for `uncertainties.testing` though. Could we use one of the options in [this StackOverflow question](https://stackoverflow.com/questions/33508060/create-and-import-helper-functions-in-tests-without-creating-packages-in-test-di). Maybe the `conftest.py` and fixture option, or the `pythonpath` and helpers file option?

wshanks (commented):
This looks good. Besides my inline question, could we add this to `pyproject.toml`:

```
[tool.pytest.ini_options]
testpaths = ["tests"]
```

I don't know that we need to add any additional options at this time.

tests/test_umath.py
jagerber48: Adding `testpaths = ["tests"]` is not required for tests to work right? It's just a convenience that makes `pytest` discover tests more quickly? Is my understanding correct?

In the past I've put helper function needed for testing in an `__init__.py` file in the `tests` folder. These objects can then be imported within the test scripts. We could also have a `tests/helper.py` module to accomplish the same result.

andrewgsavage: Yea, importing across test files does feel a bit odd. The helpers.py solution is nice and simple.

wshanks: > Adding testpaths = ["tests"] is not required for tests to work right? It's just a convenience that makes pytest discover tests more quickly? Is my understanding correct?

Yes, that seems to be all it does functionally. My conjecture was that it might also help IDE's figure out that the project uses pytest as the test runner because otherwise there is no configuration indicating that pytest should be run, but I haven't tried to confirm that.

I like the move to helpers.py. We might want to add

```
testpaths = ["tests"]
addopts = [
    "--import-mode=importlib",
]
```

to pyproject.toml like the one StackOverflow answer suggests, but we could wait for that. Currently, pytest adds all the test directories to the Python path so you can do `import helpers` but it has been trying to move away from that to just importing the test files with `importlib` without modifying the Python path.

wshanks (approved):
I think this is good. I had one suggestion about adding more pytest configuration but it's not mandatory.

wshanks:
Also, should this fold in #210? Otherwise they will create merge conflicts with each other.

tests/test_umath.py
andrewgsavage: thanks for explaining, wouldn't have understood why otherwise

andrewgsavage:
> I think this is good. I had one suggestion about adding more pytest configuration but it's not mandatory.

I tried doing this and it worked locally but I couldn't get it working with CI. I've reverted so this can be merged and we can do the importlib in another PR. Ready for merging now

wshanks (dismissed):

wshanks (approved):

wshanks:
It looks good but it won't let me merge at the moment. The diff also doesn't look right in the GitHub web UI but looks right locally. I saw this in another repo recently as well.

jagerber48:
Can `testing.py` be moved into `tests/helpers.py`?

wshanks:
> Can `testing.py` be moved into `tests/helpers.py`?

It could be, but @andrewgsavage described the contents above as "I moved some functions that could be used by downstream libraries to uncertainties.testing"

andrewgsavage:
hold up, looks like I'd deleted the contents of a file

wshanks:
I see all the changes from #214 still in the web UI. Could you squash onto master and force push? Or does it show up correctly for others now?

jagerber48:
The `testing.py` concept seems strange to me.

> I moved some functions that could be used by downstream libraries to uncertainties.testing

These functions are only used in the tests so I think it makes sense for them to be in the test helpers module. Is the idea that downstream users might want to use these functions? If so, then they should be part of public API and they need to be documented etc.  And if they're meant to be part of the public API I think it is strange for them to be in a module named `testing` instead of a module name `utilities` or something.

newville:
I am not ever going to be able to keep up with conversations at the pace here.  

I do not fully understand the added `uncertainties/testing.py` file either.   Do we want to support users using these tests? 

I am working slowly on the docs, but this will take some time -- given my work schedule, into next week.

I deeply despise that codecov calls a PR unsuccessful if it loses a bit of coverage.  Knowing the code coverage is fine. As a github check, codecov and its reports are warts.  I will make a PR to copy the codecov config from lmfit.

PR #212, opened 2024-04-01
Added .vscode folder to .gitignore settings.

willynilly:
Added .vscode folder to. gitignore settings.


jagerber48:
I always have a hard time getting the entries in `.gitignore` right. Would it be better if this were `.vscode/`? Or maybe even `.vscode/**`? `.idea/` for pycharm has worked fine for me.

newville:
@jagerber48 I think

```
.vscode/
```

which will ignore that folder and everything in it.

willynilly:
I changed it as suggested.

newville (approved):

PR #214, opened 2024-04-01
remove 1to2 and depreciations

andrewgsavage:
Here I've removed:
- the 1to2 code, 
- the compatability code that handled obselete function calls, 
- the depreciations in these calls,
- the tests that tested obselete behaviour


wshanks:
This looks good to me. Could you remove the 1to2 section from `doc/index.rst` as well?

Also, perhaps a separate discussion, but should the codecov checks be optional? It is nice to keep track of the coverage, but when it is not 100% it is hard to say whether it is good or bad that the number goes down slightly as code is added and removed. I think the ideal case is to require 100% coverage and mark code that doesn't need to be covered as "no cover".

jagerber48:
Mmm I'm not exactly sure how the codecov CI works. the github action that I set up submits the code to codecov.io, but then it seems like codecov.io sends data back as a github action that can pass or fail. 

I'm not sure how we would "make codecov checks optional". Right now I don't think they're part of the "final results" github action, so codecov not passing doesn't block merging a PR. This seems fine to me.

> I think the ideal case is to require 100% coverage and mark code that doesn't need to be covered as "no cover".

I haven't looked at the code coverage on `uncertainties` yet. This PR will help clean things up and then we'll be able to get a clear picture of if `uncertainties` is hitting 100% coverage or not. But yes, once we have that picture we might mark some lines as no cover, but I prefer to be pretty conservative when it comes to marking lines as no cover. We may need to write more tests.

andrewgsavage:
> remove the 1to2 section from `doc/index.rst`

done

I think codecov is set up fine, we can use our own judgement for cases like this and still merge

wshanks (approved):
Looks good to me now!

I think you are right about codecov already being optional in the way that I meant. In any case, it is passing now :slightly_smiling_face:

newville:
Looks great.  
Does anyone object to merging this?

wshanks:
I think we had enough consensus to merge here!

PR #221, opened 2024-05-10
pr_template

andrewgsavage:
I think this is what is needed for github to give a template when opening a PR

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,5 @@
+- [ ] Closes # (insert issue number)
+- [ ] Executed `pre-commit run --all-files` with no errors
+- [ ] The change is fully covered by automated unit tests
+- [ ] Documented in docs/ as appropriate
+- [ ] Added an entry to the CHANGES file


newville (approved):
Thanks -- that looks great.

andrewgsavage:
we should get pre-commit running first!

newville:
@andrewgsavage oh, right!

newville:
@andrewgsavage Is the file at https://github.com/lmfit/uncertainties/blob/pr221/.pre-commit-config.yaml a reasonable start?  
I forked your PR, added that and pushed it to a "pr221" branch here (I don't know how to push to an open PR from a user branch).   You could take and add that (if it looks acceptable).

wshanks:
Sorry, I haven't had as much time for Uncertainties recently. Ideally the pre-commit check is enforced by the CI, so it might not need to be in the list any way?

You might add some note for the PR author within `<!-- -->` so that it is visible while writing the PR but not in the rendered version to explain how the author should use the list -- put the PR description above/below, check things off or leave them for a reviewer?

PR #222, opened 2024-05-15
pin numpy <2

andrewgsavage:
theres some depreciations in numpy 2, so they recommend pinning to <2 until modules support /test it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ Issues = "https://github.com/lmfit/uncertainties/issues"
 Changelog = "https://github.com/lmfit/uncertainties/blob/master/CHANGES.rst"
 
 [project.optional-dependencies]
-arrays = ["numpy"]
+arrays = ["numpy<2.0"]
 test = ["pytest", "pytest_cov"]
 doc = ["sphinx", "sphinx-copybutton", "python-docs-theme"]
 all = ["uncertainties[doc,test,arrays]"]


newville (approved):

newville:
@andrewgsavage Sounds like a wise idea to me.  Any objections to merging this?

andrewgsavage:
merge when ready

wshanks:
A quick `ruff check --select=NPY` flags no compatibility issues and running the tests with `numpy==2.0.0rc2` produces only one test failure -- `test_broadcast_funcs` fails asserting that `numpy.acos` does not exist but it was added in numpy 2.0. It is probably better just to adjust that test.

newville:
@wshanks Thanks for that!  That's really good news, actually.

I would still vote for merging this PR for 3.2.0, and expect to "officially" support numpy2.0 support in the next release.  

I expect that many libraries will consider numpy 2.0.0 to be "beta", and wait until the first or second patch release before supporting it.

andrewgsavage:
Since it's just that small change it would be good to do that and have
numpy 2 support

newville:
@wshanks @andrewgsavage Thanks!  I understand that upper bounds on versions are not ideal in general and does not always work well.

Still, I might suggest that we merge both #225 and this PR to limit `numpy < 2`,  at least for now, and then be ready to lift this restriction soon.  (I am also +1 on #223 and #226).  I think we can be cautious about supporting Numpy 2 for a little while, even if we believe it will mostly work just fine.

As with #224 and the recently re-pinged Issue #114, I think there are a lot of "legacy corner cases" in the unumpy interface, - including matrix support - that may need to be carefully checked with NumPy 2.

But: I could be convinced to just go for it and allow Numpy 2 and (try to) deal with any consequences.

andrewgsavage:
I would rather not merge this since, to the best of our knowledge,
uncertainties works with numpy2.

If we do find something that doesn't work, that's a good reason to pin <2.

newville:
@andrewgsavage I am OK with closing this.  We can re-open if needed.

Issue #215, opened 2024-04-02
Changelog maintenance plan


wshanks:
It would be good to keep a change log as changes are made instead of writing it at release time. We have CHANGES.rst now. Do we just add notes to it as we go or shall we use a release notes management system? Using a system is nice for avoiding merge conflicts and keeping the notes formatted.

The systems I know about from use in other projects are the following. They all follow the pattern of holding a directory of files with notes that get committed along with code changes and then compiling the notes into the changelog file at release time.

1. [Towncrier](https://github.com/twisted/towncrier): Most popular. Used by pytest, pip, numpy, etc.
2. [Reno](https://github.com/openstack/reno): From the openstack project.
3. [Scriv](https://github.com/nedbat/scriv): From the author of coverage.py and used by coverage.py.

I would go with Towncrier since it is the most popular unless there is an argument for another option. They all seem similar to me. I have more experience with Reno, but I haven't seen anything from it to make it stand out from Towncrier.

I noted the points I would put in a new release currently here: https://github.com/lmfit/uncertainties/discussions/193#discussioncomment-8977264.

newville:
@wshanks Thanks - that sounds like a good idea to me.  I have to admit to not being familiar with any of these. The READMEs of both Towncrier and Reno make it sound like each is a really valuable tool but with slightly different aims.

+1 on the concept, I defer to you on which to use.

marwin1991:
I don't know if you are looking for such a  tool, but you can check out https://logchange.dev/tools/logchange/getting-started/

Jinja templates should meet your needs.

Apache Solr using it exactly because of wasting time for pull requests conflicts

https://lists.apache.org/thread/4dotf4qx4ss3qr3xonv2y63v7wdg40nt

PR #223, opened 2024-05-15
changes

andrewgsavage:


newville (approved):
Looks good to me.

Should we adopt a "more conventional" Changelog with a heading for each version (with date) and then lists of "bug fixes" and "new features"?  I think that was suggested at some point.... But I also get that it can be some work ;)

wshanks:
I had suggested something more systematic in #215 but I have not worked on it so far.

andrewgsavage:
That would be preferable yes. This was just so we can get a release out.

A simple list, one line per PR to ensure it's continually up to date would
work too.

PR #225, opened 2024-05-16
Update unumpy test to be compatible with numpy 2

wshanks:
`uncertainties` defines error-propagating versions of several of the functions in the `math` module. Then in the `unumpy` code, it iterates over this same set and defines array-compatible versions of those functions using `numpy.vectorize`. Because before numpy 2 numpy used different names for the inverse trigonometric functions (`arccos` vs `acos`), the code needed to change the name for the numpy version. One place in the tests just checked that `arccos` was in the numpy module and `acos` was not. Since numpy 2 added the aliases for the `a`-named versions of the `arc` functions, that test fails with numpy 2. Since `uncertainties` does not really need to test the names of functions in `numpy` that check is removed.

A future update could add the `acos()`-style aliases to `unumpy` as well.


newville (approved):
Thanks, looks good to me.

PR #226, opened 2024-05-16
Mark docstrings with backslashes as raw strings

wshanks:
This change avoids Python issuing a SyntaxWarning about an unrecognized escape sequence. The only occurrences were in one test file.

Note that some strings had a `u` prefix replaced with `r`. The `u`
prefix was needed for Python 2. These strings did not work with a `ur`
prefix. Since Python 2 is no longer supported, no other means of keeping
support with Python 2 was attempted.


newville (approved):
Thanks!

newville:
@wshanks I am +1 on just merging this.  Any objections?

wshanks:
> @wshanks I am +1 on just merging this. Any objections?

Not from me :slightly_smiling_face:

PR #228, opened 2024-05-30
add global TOC to doc index page

newville:
This adds a Table of Contents to the sidebar for the index page of the docs. 
See  https://lmfit.github.io/uncertainties/ for results


wshanks (approved):
This is an improvement to me. Especially on mobile (or just making the browser window narrow), it is nice to have the navigating on the side.

PR #230, opened 2024-06-02
3.2.0release

newville:
We need to set the version in `pyproject.toml` too!  

Also, apparently `uncertainties.unumpy` was not being installed  -- how did tests pass?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "uncertainties"
-version = "3.1.7"
+version = "3.2.0"
 authors = [
     {name = "Eric O. LEBIGOT (EOL)", email = "eric.lebigot@normalesup.org"},
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ classifiers = [
 dependencies = []
 
 [tool.setuptools.packages.find]
-include = ["uncertainties"]
+include = ["uncertainties", "uncertainties.unumpy"]
 exclude = ["doc", "test"]
 
 [project.urls]


wshanks:
Hmm, `tool.setuptools.packages.find` should recurse into sub packages. I think I compared the wheel file lists before and after the pyproject.toml change and they agreed (I can't check at the moment).

Regarding the tests, we don't test the packaging currently because the uncertainties directory is top level and we run the tests from the top level. So the tests would pass even if we did not install the package. We could adjust things to make sure the installed files are tested.

newville:
In a fresh devel environment (deleting the `build` folder),  `pip install .` does not install the `unumpy` submodule.  And then running the tests by hand failed.

wshanks:
Could it be that you just did not have numpy installed because you did not do `pip install .[arrays]`? For me, `pip install .` does install `uncertainties/unumpy` in the `site-packages` directory and `python -m build .` does produce a wheel with `uncertainties/unumpy`.

newville:
I do have `numpy` installed.   

```
 > rm -rf build 
 > rm -rf  $CONDA_PREFIX/lib/python3.11/site-packages/uncertainties*
 > pip install numpy
 > pip install ".[all]"
 > ls   $CONDA_PREFIX/lib/python3.11/site-packages/uncertainties
__init__.py   __pycache__   core.py       umath.py      umath_core.py
```

`uncertainties.unumpy` is not installed.

You could use `pyproject.toml` with


```
[tool.setuptools.packages.find]
include = ["uncertainties*"]
exclude = ["doc", "test"]
```

to find subpackages, but 

```
[tool.setuptools.packages.find]
include = ["uncertainties"]
```

will not find subpackages.

wshanks (approved):
Ah, you're right! I hadn't tried cleaning my repository before. It must have only worked for me because of the `build/` directory or some other left over metadata.

Reading the setuptools documentation again, I think it would be better to change `include` to `uncertainties*` and get rid of `exclude` but what is in this PR is fine.

newville:
@wshanks Thanks.   In other projects, I have favored an explicit list of sub-packages to avoid mistakes.  But, now that "tests" have moved out, I think "uncertainties*" would be OK.

wshanks:
This is such a small project that an explicit list seems fine. I will open a PR to ensure that the package as installed from the wheel is tested in CI. I can switch to an explicit list (without `find`) there.

wshanks:
I opened https://github.com/lmfit/uncertainties/pull/232 to make sure we test the installed version of the package.

PR #231, opened 2024-06-03
Ensure that GitHub Actions tests against the installed package

wshanks:
PYTHONSAFEPATH=1 will prevent Python from loading the uncertainties files from the repo directory instead of loading them from site-packages (at least for Python >=3.11 when PYTHONSAFEPATH was added).

pytest's `import-mode=importlib` will prevent `pytest` from adding the working directory to the Python path.


wshanks:
Sorry, I mean to open this one on my fork for testing.

PR #232, opened 2024-06-03
Use explicit package list and ensure it is tested in CI

wshanks:
* In `pyproject.toml`, use an explicit package list instead of using the `find` directive.

* In GitHub Actions workflow, do not use an editable install since that implicitly includes all modules below `uncertainties/` and we want to make sure we test the package in its installed form.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -28,8 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


newville (approved):
Thanks!

wshanks:
Hmm, it looks like this caused the way we have coverage set up not to count the coverage of the main package's code. I will look into getting that counted while still testing the installed package. Anyone else is welcome to look into it as well.

Issue #234, opened 2024-06-03
`tests/helpers.py` is missing in 3.2.0 PyPI sdist


mgorny:
When running the test suite from PyPI sdist for 3.2.0, I'm getting errors like the following:

```pytb
_______________________________________________ ERROR collecting tests/test_ulinalg.py ________________________________________________
ImportError while importing test module '/tmp/portage/dev-python/uncertainties-3.2.0/work/uncertainties-3.2.0/tests/test_ulinalg.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.10/site-packages/_pytest/python.py:492: in importtestmodule
    mod = import_path(
/usr/lib/python3.10/site-packages/_pytest/pathlib.py:591: in import_path
    importlib.import_module(module_name)
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1050: in _gcd_import
    ???
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
/usr/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:178: in exec_module
    exec(co, module.__dict__)
tests/test_ulinalg.py:11: in <module>
    from helpers import uarrays_close
E   ModuleNotFoundError: No module named 'helpers'
```

Indeed, `tests/helpers.py` is not included in the sdist.

wshanks:
Sorry about that. I opened #235 for this.

#235 should be enough. I opened #232 to make sure that the installed package files are tested and not the files in the repo. We could go further and make sure the files from the sdist are used for testing as well :thinking: I haven't noticed other projects going that far though.

newville:
@mgorny Yeah, thanks and sorry for the trouble.  It looks like the building of the ".tar.gz" file included many of the files in "tests", but not the "helpers.py" file.   I'm not sure why that is.  

We're finding a few packaging- and release-related buglets.
I suspect that we will need to release a 3.2.1 relatively soon to fix these kinds of problems.

PR #235, opened 2024-06-03
Use setuptools-scm to ensure all repo files are in sdist

wshanks:
Listing `setuptools_scm` in `pyproject.toml` enables it to be used to determine which files are included in the sdist instead of the older `MANIFEST.in` procedure. `setuptools_scm` just includes every file tracked by `git` which seems fine for this small project.

Additionally, delete `INSTALL.txt` which had outdated installation information.


wshanks:
Closes #234

mgorny:
Thanks. I can confirm that with this change, `python -m build -s` includes all files in sdist.

newville:
@wshanks Thanks -- we could also use this to set the version number from the git tag, as will

- `pyproject.toml`:
add
```
[tool.setuptools_scm]
write_to = "uncertainties/version.py"
version_scheme = "post-release"
```
and remove

```
version = "3.2.0"
```


- `uncertainties/__init__.py`:

replace 
```
# Numerical version:
__version_info__ = (3, 2, 0)
__version__ = '.'.join(map(str, __version_info__))

```

with

```
from .version import version as __version__
```

It would be OK with me to postpone this until later.

newville (approved):

wshanks:
I like the idea of using `setuptools-scm` to manage the version. I will look into doing that in a later PR.

PR #236, opened 2024-06-04
Readthedocs config more

newville:
more (maybe last? ;)) updates to readthedocs configs, hoping it now connects and builds docs.


andrewgsavage (approved):

andrewgsavage:
docs still didn't build... there should be a way to get rtd to build pull requests too

newville:
@andrewgsavage Thanks -- I agree that rtd should be able to run on PRs.  I'm looking into that (and the actual error).

I do not have much experience with rtd, but I'm not loving it over Github Pages.

PR #246, opened 2024-07-04
Codecov updates

newville:
This updates the `.codecov.yml` file to be (I believe!) more compliant with codecov expectations.
And it changes the invocation of pytest with coverage in the testing environment.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Test source code and docs
       run: |
         cd tests
-        python -m pytest --cov=uncertainties --cov-report=xml
+        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.0.1
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -31,7 +31,8 @@ jobs:
         python -m pip install .[arrays,test]
     - name: Test source code and docs
       run: |
-        pytest --cov . --cov-report xml
+        cd tests
+        python -m pytest --cov=uncertainties --cov-report=xml
```
wshanks, line 35: This nicely fixes what I broke when I made sure that the installed package was tested instead of the code in the repo. However, it excludes tracking coverage of the test files. We can track them as well by using a second `--cov`. I also think it is convenient to add `--cov-report=term` so that the percentages can be seen in the GitHub Actions output.

```suggestion
        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
```

wshanks (approved):
Looks good!

newville:
@wshanks Thanks.  If there are no objections (@andrewsavage  @jagerber48), I suggest you squash and merge this. We could then tag and release 3.2.2, and push to PyPI.
   * fixing support for Numpy2.
   * showing more realistic coverage stats.

andrewgsavage:
nice, go for it

PR #248, opened 2024-07-11
pre-commit linting

andrewgsavage:


andrewgsavage:
Most of the diff standardises formatting which it changed automatically. There were a few things which it didn't fix automatically with obvious fixes that I fixed, anything not so obvious I've told the linter to ignore using `# noqa`

wshanks (approved):
This is great, @andrewgsavage. I would like to remove the `# noqa`s by addressing them, but already what is here is an improvement, so I think we can merge this big change for now and come back to the individual linter issues in follow up PRs.

newville:
+1 from me.  Also, we should ALWAYS ignore alleged failures from codecov.  A change in coverage score is not ever a failure that prevents merging.  codecov is very, very wrong here.

wshanks:
Yes, @newville, I had hoped that #246 was going to remove the :x: from Codecov on reasonable PRs not affecting coverage much, but it seems like more adjustments to the configuration are needed.

PR #250, opened 2024-07-12
Remove some <py3.3 code

andrewgsavage:

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -15,7 +15,7 @@ Main module for the uncertainties package, with internal functions.
 from __future__ import division  # Many analytical derivatives depend on this
 
 from builtins import str, zip, range, object
-from math import sqrt, isnan, isinf  # Optimization: no attribute look-up
+from math import sqrt, isfinite  # Optimization: no attribute look-up
 
 import copy
 import collections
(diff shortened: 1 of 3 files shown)


uncertainties/core.py
```diff
@@ -790,7 +782,7 @@ def std_dev(self, std_dev):
         # (Note: if NaN < 0 is False, there is no need to test
         # separately for NaN. But this is not guaranteed, even if it
         # should work on most platforms.)
-        if std_dev < 0 and not isinfinite(std_dev):
+        if std_dev < 0 and not (isinf(std_dev) or isnan(std_dev)):
```
wshanks, line 785: I see that this maintains the current behavior because `math.isinfinite` does not exist. Should we take the opportunity here to do what the code was trying to do like this?

```suggestion
        if std_dev < 0 and isfinite(std_dev):
```

with a `from math import isfinite` above? I think previously `isinfinite` may have been accidentally added instead of `isfinite`.

wshanks (approved):
This looks fine to me. I had one question, but I am fine with merging if you prefer the way the code is now.

wshanks (approved):

newville:
+1 from me!

PR #252, opened 2024-07-13
Add CI run that tests the package without numpy installed

wshanks:
Given that it is common to work with Uncertainties and numpy, it is easy to forget to check that a hard dependency on numpy has been introduced. Here an extra run is added to the CI configuration to run the tests without installing numpy.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -38,6 +38,29 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         slug: lmfit/uncertainties
+  test_without_numpy:
+    name: Test without numpy
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python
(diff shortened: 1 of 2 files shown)


wshanks:
Oops, I had meant to open this against my fork first for iterating on the yaml syntax. It looks like it is good now.

At first, I experimented with adding a `--skip-numpy` option to pytest and marking all the numpy dependent tests so they would show up as skipped, but pytest tries to import all the test files as part of its collection so for this to work a lot of numpy related imports had to be put in try blocks. I didn't like how that would make numpy installation problems show up instead as "variable numpy is undefined" so I went with this simpler variant of just ignoring entirely the files that import numpy (and skipping on test in test_uncertainties that tries to import numpy).

andrewgsavage (approved):

andrewgsavage:
nice this will be helpful!

wshanks:
I think this one is simple enough to merge without more feedback (it doesn't affect the package itself).

PR #255, opened 2024-07-14
have readthedocs install with , drop requirements files

### LINKED ISSUE #254: readthedocs docs not building
The automated build of the docs at readthedocs.org is failing, for example https://readthedocs.org/api/v2/build/24927988.txt
The readthedocs version of the docs is stuck on version 3.2.1.

That build log suggests that it is fetching uncertainties from GitHub and installing stuff in `doc/requirements.txt`, but maybe not actually installing it with `pip install`, and then complaining that the file `uncertainties/version.py` does not exist.   Well, that file is not in the repository and is generated at `pip install` time.  Hm...

newville:
- [x] Closes #254

This changes the readthedocs config to install uncertainties and dependencies with `pip install ".[all]"`.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -29,4 +29,4 @@ sphinx:
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
 python:
   install:
-  - requirements: doc/requirements.txt
+  - pip install ".[all]"
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,0 @@
-sphinx
-sphinx-copybutton
-python-docs-theme
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,0 @@
-numpy
-pytest
-pytest-cov
-sphinx


jagerber48:
This is my first time checking out readthedocs config. On the readthedocs page for `lmfit/uncertainties` I see the builds are failing. And I don't see the build for this branch. I can't see the settings for readthedocs. Is there a different readthedocs page somewhere? I see lebigot as a maintainer but no one else. Could others be added as maintainers? @newville, you've done a lot of docs work, have you been able to access readthedocs settings?

newville:
@jagerber48 
> This is my first time checking out readthedocs config. On the readthedocs page for lmfit/uncertainties I see the builds are failing. And I don't see the build for this branch. I can't see the settings for readthedocs. 
> Is there a different readthedocs page somewhere? 

Not that I am aware of. 

> I see lebigot as a maintainer but no one else. Could others be added as maintainers? @newville, you've done a lot of docs work, have you been able to access readthedocs settings?

No, I do not.

jagerber48:
Ok got it. There's `uncertainties.readthedocs` and legacy `uncertainties-python-package.readthedocs`. I'll ignore the latter, though it does show up first on google etc.

Regarding readthedocs installation issues. 

- I worry that this will cause readthedocs to try to `pip install uncertainties` from `pypi` rather than pip installing the local version.
- I worry that readthedocs will lack the sphinx packages it needs which are currently found in the `requirements.txt` file that it uses

My recommended fix is to add `.` to the `requirements.txt` file so that the readthedocs build runs `pip install .` in its local `uncertainties` directory. When it does this `uncertainties` will get built, including the package info file which is needed for reading the version using `importlib.metadata.version` in `conf.py`. I have this setup at https://github.com/lmfit/uncertainties/pull/257

I don't know if this is the "best" strategy but it works for `sciform`.

jagerber48:
Ah, I see the doc requirements are included in `.[all]`. This fix is good then.

.readthedocs.yml
```diff
@@ -29,4 +29,4 @@ sphinx:
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
 python:
   install:
-  - requirements: doc/requirements.txt
+  - pip install ".[all]"
```
jagerber48, line 32: How about `pip install ".[doc]"` or `python -m pip install ".[doc]"`? No need to spend build resources installing `numpy` when sphinx should only need the doc building requirements.

Also this line could be indented two spaces I think?

newville: a) let's allow the assumption that numpy and pytest are installed when building the docs, in case we or anyone else wants to eventually dynamically create content (say, a gallery or jupyter directives).  

b) It was not indented before, and I am not 100% sure what rtd expects.

c) It is not that I like rtd even a tiny amount, just trying to get the d*n thing to not fail.  I am very much in favor of dropping rtd for GitHub pages.

jagerber48 (approved):

jagerber48:
I’d like to see this run on rtd before merging but I think we would need admin privileges. Better just get this in and see if it works. @newville I’ve missed the etiquette we’ve been following for PRs lately. When would be appropriate for me or someone to merge this in?

PR #262, opened 2024-08-12
Eliminate the `Variable` class

jagerber48:

This PR:
- Eliminates the `Variable` class and eliminates the `AffineScalarFunc.derivatives` property. 
- The `AffineScalarFunc` class is renamed to `UFloat` with `AffineScalarFunc` left as a legacy name.
- The `LinearCombo` class is refactored into the `UCombo` class but the architectural role is similar: track the uncertainties of `UFloat` objects as a lazily expanded combination of objects with uncertainty
- The `UAtom` object is introduced as the fundamental element which captures uncertainty. Each `UAtom` models a random variable with zero mean and unity standard deviation/variance. Each `UAtom` has a `uuid` and all `UAtom` are mutually statistically independent. The expanded version of the `UCombo` maps `dict[UAtom, float]`.

For more detail/discussion see https://github.com/lmfit/uncertainties/discussions/251#discussioncomment-10309160

Note that as of the time opening this PR the PR is still a work in progress. Not all elements in the bullet list above are implemented and  a good handful of tests are still failing.


tests/test_uncertainties.py
```diff
@@ -104,9 +105,6 @@ def test_ufloat_fromstr():
         "3.4(nan)e10": (3.4e10, float("nan")),
         # NaN value:
         "nan+/-3.14e2": (float("nan"), 314),
-        # "Double-floats"
```
jagerber48, line 107: These floats are too large to work with. Prior to this PR the large float value could be saved into the `AffineScalarFunc` object and displayed, but  any arithmetic with such large values would result in an `OverflowError`.  In the new framework even displaying a `UFloat` the first time requires a calculation, even if it's as simple as `sqrt(x**2)`. So in the new framework the `OverflowError` appears immediately.

jagerber48:
All the tests are passing except some having to do with the uarray inverse and pseudo inverse functions. The code for this is pretty complicated. It might make sense to just rewrite the code to convert array-compatible functions to ufloat array compatible functions.

tests/test_unumpy.py
```diff
@@ -179,7 +162,7 @@ def test_wrap_array_func():
     # Function that works with numbers with uncertainties in mat (if
     # mat is an uncertainties.unumpy.matrix):
     def f_unc(mat, *args, **kwargs):
-        return mat.I + args[0] * kwargs["factor"]
+        return numpy.matrix(mat).I + args[0] * kwargs["factor"]
```
jagerber48, line 165: This function needs to cast to `numpy.matrix` internally. This is because `to_uarray_func` currently drops compatibility with `numpy.matrix`.

jagerber48:
Prior to commit `c80dd9b` performance was terrible because on each math operation I was calculating the `std_dev` of one of the operands. But calculating `std_dev` (specifically expanding the `UCombo`) is exactly the thing which must be avoided to get the performance enhancement from the lazy evaluation. after this change and some quick benchmarking it is looking like this branch is performing similarly to `master` on the 

```
str(sum(ufloat(1, 1) for _ in range(10000)))
```
benchmark.

____________
Before merging in this PR I am going to make a new one which includes benchmarking tests. I'm not exactly sure how to code up benchmarking tests. but I'll code up some ideas and we can discuss in that PR.

jagerber48:
Commits `5fc3cbd` and `6d53afb` introduces some code that special cases situations in which you have an empty uncertainty or an uncertainty with a zero weighting somewhere. These special cases were needed to ensure that something like

```
from uncertainties.umath import pow

negative = ufloat(-0.3, 0.01)
integer = ufloat(-3, 0)

pow(negative, integer)
```
gives back the same as
```
pow(negative, -3)
```
Without this special casing you end up with some kind of `nan` in the mix. Basically the special cases in those commits support the model that "a `UFloat` with zero std_dev is equivalent to a regular `float`.

Is this truly the behavior we want? Do we want these special cases?

uncertainties/ucombo.py
```diff
@@ -0,0 +1,134 @@
...
+        self._hash = hash(self.uuid)
+
+    def __eq__(self: UAtom, other: UAtom) -> bool:
+        return hash(self) == hash(other)
+
+    def __hash__(self: UAtom) -> int:
+        return self._hash
+
+    def __lt__(self: UAtom, other: UAtom) -> bool:
+        return self.uuid < other.uuid
```
andrewgsavage, line 25: codecov gives a load of warnings for UAtom and UCombo. Would bit be worth having tests to show methods like eq, lt behave as expected?

it's also a bit suprising some of these lines (eg for UCombo's eq) don't get hit by any of the existing tests

jagerber48: Thanks for taking a look. I haven't looked at codecov yet. The story for why `UCombo`s `__eq__` doesn't get hit is worth sharing.

First off, before this PR and as of this comment, `UFloat.__eq__` does not check for equality of the nominal value and equality of the uncertainty as I would have naively implemented. Rather, it looks at the difference between `self` and `other` and checks that the resulting `UFloat` has zero nominal value and zero std_dev. Because subtraction is supported between `UFloat` and `float`, this allows `__eq__` to easily support comparison between `UFloat` and `float`. In going through the tests I've learned that, for better or for worse, `uncertainties` has definitely chosen the convention that a `UFloat` with a `std_dev` of zero should behave like a regular `float`.

I am still using this difference calculation for `UFloat.__eq__`. It would be more natural to me to have `__eq__` be something like
```
def __eq__(self, other):
    if isinstance(other, UFloat):
        return (self.nominal_value == other.nominal_value and self.uncertainty = other.uncertainty)
    else:
        if self.std_dev != 0:
            return False
        else:
            return self.nominal_value == other
```
But currently the code implementing `__eq__` is defined in `ops.py` and gets monkey patched onto `UFloat`. I can't import `UFloat` into this function because it would create circular imports between `ops.py` and `core.py`.

Anyways, I can look into disentangling this.

________________
Note though, that I tried something like the above code for a second and found out that the current version of `UCombo.__eq__` was broken because I didn't know `sorted` returned a list! Easy fix, but emphasizes the importance of testing.

Note that I made the decision that `UCombo` `__eq__` and `__hash__` use the `expanded` dictionary for comparison. Thuis means that two `UCombo`s with different `ucombo_tuple`s could compare equal. For `UFloat` I think it is clear that we want two `UFloat`s to compare based on the expanded dictionary from the uncertainty. For `UCombo` it wasn't as obvious to me. For example, if `dx` is a `UAtom` then it would be possible for there to be a nested `UCombo` storing something like
```
dx + (-dx + (dx + (-dx + (dx + (-dx + ...)))))
```
Which ultimately is equal to 0. But the former might take a longer time to compute. 

Actually I think I need to add a special case for 0 so that the presence of a `UAtom` in one `UCombo` expanded dict but not the other doesn't matter if the weight is equal to 0.

In any case, does this make sense to allow two `UCombo` to compare the same even if they have different `ucombo_tuple`? Or should that distinction be made apparent in `__eq__`  and `__hash__` (but `UFloat` will still compare based on `expanded` dict)?

jagerber48:
Ok, it is time to pick the names of the classes because I want to start updating the docs (I'm not technically blocked but would be nice to get it right earlier). Right now we have `UFloat`, UCombo` and `UAtom`. These classes are modeling the following.

For reference, suppose we have
```
x = x_0 + dx
y = y_0 + dy
z = x*y = x_0 * y_0 + (y_0 * dx + x_0 * dy)
```
`dx` and `dy` are `UAtom`s (individual random elements with zero mean and unity standard deviation/variance), `1*dx`, `1*dy`, and `y_0*dx + x_0*dy` are `UCombo`s (float-weighted linear combinations of `UAtom`s) and `x`, `y`, and `z` are `UFloat`s (nominal value + uncertainty given as a `UCombo`).

Notes on names:

- `UFloat`: I think we're all in agreement that this name is fine. `AffineScalarFunc` will remain a legacy alias that I suggest we add a `FutureWarning` too in an upcoming release (we shouldn't do that in this PR however since I don't want to get bogged down in that discussion yet. For the record, I will say there has been discussion about the name `UNumber` for something that would support more types of nominal values and uncertainties than just `floats`. E.g. `Decimal` or possibly complex numbers or something. But given `UFloat` is already in use I say it makes sense to keep it.
- `UCombo`: This class roughly replaces the old `LinearCombination` class. I don't have strong opinions on this name. `LinearCombination` would be fine, but I would like to somehow emphasize that is a linear combination **of zero-mean random variables** which is not captured by just `LinearCombination`. Other obvious options: `ULinCombo`, `ULinearCombo`, `ULinearCombination`. The same options without a `U`. Again, my opinions aren't strong here. If I don't hear otherwise I will keep `UCombo`.
- `UAtom`. I think this is the most controversial. `UAtom` kind of serves to replace the old `Variable` class, but it is much simpler than the `Variable` class and this reduction in complexity is what drives the overall improvement in the architectures (IMO). I chose the name `UAtom` because it is an atomic/irreducible/indivisible unit (this is what is meant by "atom") of uncertainty. A `UAtom` has zero mean, unity variance, and is not correlated with any other `UAtom`. This is in contrast with the `UCombo` which is a combination of many `UAtom`. However, it's been pointed out that users outside of the physical sciences may be less familiar with the term "Atom" and what it might mean here. Open to proposals for alternative names. I don't have good ideas. `UnitUncertainty`, `StdNormUnit` (standing for "standard normal unit" referring to the standard normal distribution), `RandomUnit`, `URandom`. I prefer `UAtom` to all of these but open to other proposals and discussion.

jagerber48:
@newville sorry, this PR originally targeted `master` but with our plan to release `4.x` from `main` this PR should merge into `main`. I've changed the PR base just now. 

I think merging this PR into `main` will be the big PR review before we release `4.0.0`. This PR is not ready for review. It still needs

(1) to address the performance regression. Before I would say it had a major performance regression. Now the regression is only ~20%. At least on the one performance benchmark we have. See #327.
(2) The documentation needs to be rewritten taking into account that new changes that is what I'm working on presently
(3) During and after the above two items I will be going over things myself and I imagine there will continue to be small changes to the source code and tests.

uncertainties/ucombo.py
```diff
@@ -0,0 +1,141 @@
...
+import random
+from typing import Optional, Tuple, Union
+
+
+class UAtom:
+    __slots__ = ["uuid", "tag", "_hash"]
+
+    def __init__(self: UAtom, tag: Optional[str] = None):
+        self.tag = tag
+        self.uuid: int = random.getrandbits(64)
```
wshanks, line 14: I just wanted to note how Python ints work in practice. They are variable in size. There is a 28 byte overhead for all `int`s. Beyond that, every 30 bits in size adds 4 more bytes. So getting 64 bits will give 36 byte objects (except when you randomly get a number smaller than 2**59). Maybe it would be better to do `getrandbits(60)` here so the overhead per `UAtom` is 32 bytes instead of 36 if 60 bits seems like enough randomness. I wondered if `randbytes()` would be more efficient but `bytes` has an overhead of 34 bytes instead of the 28 for int. bytes scales linearly in the number of bytes stored but remains less efficient than int at these small sizes (8 bytes / 64 bits would be 42 bytes total). A float would take only 24 bytes but I worry about populating those bytes in an unbiased way.

Issue #38, opened 2016-01-31
combined uncertainty from multiple independent measurements?


denfromufa:
Can this package handle uncertainty calculations like this?

http://physics.stackexchange.com/questions/57317/multiple-measurements-of-the-same-quantity-combining-uncertainties

If not, why?

lebigot:
The `uncertainties` package is meant to handle uncertainties like these, and even does this transparently when the uncertainties are correlated. In the particular case of the question, uncertainties would exactly give you the quoted uncertainty. If you apply the formula that transforms the measurements (with uncertainty) into the best (i.e. minimal chi-squared) estimate, you automatically get the uncertainty on the best estimate.

Does this answer your question?

denfromufa:
@lebigot yes and no :)

is there any example I can follow for this particular combined uncertainty?

lebigot:
Maybe I don't understand the question: it looks to me like everything is explained in the [documentation](http://pythonhosted.org/uncertainties/), so I am not sure what the blocking point is. Let me have a stab at it, though: the question on Stack Overflow considers a few random variables:

```
>>> from uncertainties import ufloat
>>> v1 = ufloat(1, 0.1)
>>> v2 = ufloat(2, 0.2)
>>> v3 = ufloat(1.5, 0.3)
```
If you calculate their average, `uncertainties` automatically calculates the uncertainty on the average:
```
>>> (v1+v2+v3)/3
1.5+/-0.12472191289246472
```
You can check that the formula from the question gives the same result—which you don't have to calculate, thanks to `uncertainties`.

`uncertainties` even transparently handles things like `v1-v1`, which is always equal to 0, with no uncertainty (in this case the formula from the question does not apply, but `uncertainties` does the right thing):
```
>>> v1-v1
0.0+/-0
```

Does this clear things up?

denfromufa:
Yes, definitely! Now let me look at this closer myself.

lebigot:
PS: The formula for the uncertainty in the question is incorrect: the denominator should be N^2, not N.

denfromufa:
What happens when uncorrelated uncertainties are combined?

denfromufa:
Also your example does not seem correct, the combined value is not arithmetic average, the weightss of sum are based on corresponding uncertainties.

lebigot:
[...]

PR #265, opened 2024-10-21
unumpy.average: init

doronbehar:
- [x] Addresses (already closed) #38
- [x] Executed `ruff check` with no errors related to my changes, and ran `ruff format` on the changed files.
- [ ] The change is fully covered by automated unit tests (Not sure how to do that without simply duplicating the code of the new `unumpy.average` function.


uncertainties/unumpy/core.py
```diff
@@ -30,10 +30,87 @@
...
+    # To hold the product of the dimensions to flatten
+    flatten_size = 1
+    for i in range(len(arr.shape)):
+        if i in axes:
+            flatten_size *= arr.shape[i]  # Multiply dimensions to flatten
+        else:
+            new_shape.append(arr.shape[i])  # Keep the dimension
+    # This way the shapes to average over are flattend, in the end.
+    new_shape.append(flatten_size)
+    return arr.reshape(*new_shape)
```
doronbehar, line 75: TBH, I'm not sure this does the correct thing when the original shape is something like `(4,4,4,4,4)` and `axes` is something like `(1,3)`. Suggestions are more then welcome.

doronbehar: > TBH, I'm not sure this does the correct thing when the original shape is something like `(4,4,4,4,4)` and `axes` is something like `(1,3)`. Suggestions are more then welcome.

I fixed that using `numpy.apply_along_axis` for every axis recursively.

uncertainties/unumpy/core.py
```diff
@@ -30,10 +30,73 @@
...
+            f"Cannot average over an inexistent axis {max(axes)} >= arr.ndim = "
+            f"{arr.ndim}"
+        )
+    if not isinstance(arr.flat[0], core.Variable):
+        raise ValueError(
+            "unumpy.average is meant to operate upon numpy arrays of ufloats, "
+            "not pure numpy arrays"
+        )
+    for axis in sorted(axes, reverse=True):
+        arr = numpy.apply_over_axis(_average, axis, arr)
```
doronbehar, line 96: Hmm on a 2nd thought, perhaps this means that e.g for `A` with `shape=(3,4,5,7)`, and `axes=(1,3)`, if there are correlations between `A[:,0,:,0]` and `A[:,1,:,1]` are not taken into account? Because once `axis=1` is averaged, the `covariance_matrix` call for the 1d slices of axis 1 won't take into account the correlations to the values already averaged on axis 3...

I think we can live with that, but perhaps warn the users about it in the function's __doc__, or elsewhere. Unless of course someone here will think of a better way to implement this.

jagerber48:
It sounds like you are writing code to extract the standard error on the mean of a sequence of `UFloat`s. Or maybe more generally the maximum likelihood estimator of the mean in case the individual observations are correlated? Is that correct? If so this does sound useful to me.

Maybe consider first writing code that does this and then consider extending to taking means along slices of arrays of `UFloat`'s after that? Perhaps the former function can somehow naturally be broadcast or vectorized into the latter.

doronbehar:
Hello and thanks for joining the discussion!

> It sounds like you are writing code to extract the standard error on the mean of a sequence of `UFloat`s. Or maybe more generally the maximum likelihood estimator of the mean in case the individual observations are correlated? Is that correct? If so this does sound useful to me.

Correct. 

> Maybe consider first writing code that does this and then consider extending to taking means along slices of arrays of `UFloat`'s after that? Perhaps the former function can somehow naturally be broadcast or vectorized into the latter.

That's what I did eventually in the last attempt, but I'm pretty sure that correlations between values in different axes are not taken into consideration in that case (see [my comment above](https://github.com/lmfit/uncertainties/pull/265#discussion_r1815750208)). I reworded the function's `__doc__` and added an implementation NOTE comment for maintainers.

jagerber48:
I would be more open to reviewing this idea if it did not interface with `numpy`. That is, if you provided a function that takes a collection of `UFloat` and returns a new `UFloat` calculated according to [inverse variance weighting](https://en.wikipedia.org/wiki/Inverse-variance_weighting) but which does not rely on `numpy`. 

See #253. I would say that "how `uncertainties` handles `numpy` interoperability" is up-in-the-air at the moment. I think it would be pretty easy to get this wrong. Especially if you've already been facing challenges getting things to work along the right axes, or to work according to broadcasting etc.

Also see #282 and the discussion in #277 since the discussion there may impact the fate of this PR.

PR #269, opened 2024-11-13
Make codecov pull requests informational only

### LINKED ISSUE #270: Fix various code coverage issues
There are a number of issues with the existing codecov configuration that lead to frustrations during unrelated PRs

- The main issue is the minor code coverage changes are resulting in distracting red X's showing up on PRs that should otherwise go through quickly. There are a number of causes for this
- An invalid `.codecov.yaml` file makes it so that configuration meant to loosen test failing conditions
- Codecov can be made to be purely information so that it never generates red X's
- The codecov secret token was made publicly available in the source code
- No flags on distinct uploads

jagerber48:
- [x] Closes #270 

This PR hopes to resolve a few issues with `uncertainty` package interaction with `codecov.io`. 

- #203 introduced `.codecov.yaml`. Unfortunately, I think this was an invalid configuration. Specifically because the `comment` section is supposed to be a top-level section, not nested under the `codecov` section. This means that since that PR the yaml configuration has NOT been used for our codecov updates. I think this explains why we're seeing codecov failures despite best attempts to suppress them.
- See for example #268 where a codecov failure is creating an X when we really just want the PR to go through. This PR will (attempt to) make it so that codecov coverage is reported, but will never result in a failed coverage run. This way reviewers can see if patch coverage is < 100% or if overall coverage drops due to a PR. If reviewers deem a lack or drop of coverage to be significant they can block PR approval. Otherwise, for minor coverage drops, the coverage can be easily ignored. It does so using the `informational` flag in the `.codecov.yaml` file.
- It moves the API token back to using the github secret rather than exposing it in the `.codecov.yaml` file.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -36,7 +36,6 @@ jobs:
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
-        name: ${{ matrix.os }}-${{ matrix.python-version }}
         flags: ${{ matrix.os }}-${{ matrix.python-version }}
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
@@ -61,7 +60,6 @@ jobs:
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
-        name: no-numpy
         flags: no-numpy
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
-        flags: ${{ matrix.python-version }}, ${{ matrix.os }}
+        flags: ${{ matrix.os }}-${{ matrix.python-version }}
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   test_without_numpy:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -35,6 +35,8 @@ jobs:
         python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
+      with:
+        flags: ${{ matrix.python-version }}, ${{ matrix.os }}
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   test_without_numpy:
@@ -45,7 +47,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.12
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
@@ -57,6 +59,8 @@ jobs:
           --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
+      with:
+        flags: no_numpy
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   results:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,7 +31,8 @@ jobs:
         python -m pip install .[arrays,test]
     - name: Test source code and docs
       run: |
-        python -m pytest --cov --cov-report xml --cov-report term
+        cd tests
+        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
@@ -53,8 +54,9 @@ jobs:
         python -m pip install .[test]
     - name: Test source code and docs
       run:
-        python -m pytest --ignore tests/test_unumpy.py --ignore tests/test_ulinalg.py -k "not test_monte_carlo_comparison"
-          --cov --cov-report xml --cov-report term
+        cd tests;
+        python -m pytest --ignore=test_unumpy.py --ignore=test_ulinalg.py -k "not test_monte_carlo_comparison"
+          --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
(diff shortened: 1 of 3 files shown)


jagerber48:
This PR is going very badly. I've discovered

- If I include the codecov token explicitly in the source code (something which we shouldn't do because I think it's a security risk, I think that token is something like a password) then all the 16 codecov uploads work in github actions but I don't see any uploads appear on codecov.io.
- If the codecov token is kept secret (as it should be) then all codecov upload jobs, or all but one, will fail in github actions. Maybe the one that succeeds will appear on codecov.io.

I think the issue is related to this being a PR from a forked branch. I'm not sure what needs to be done about it though.

.github/workflows/python-package.yml
```diff
@@ -53,14 +53,15 @@ jobs:
         python -m pip install .[test]
     - name: Test source code and docs
       run:
-        cd tests;
```
wshanks, line 56: Do you have to delete `cd tests;`? With the default configuration, running `pytest` puts the working directory on the Python import path and so `import uncertainties` uses the files in the repo instead of the ones installed in site-packages. If there is an error in the packaging configuration such that a necessary file is not included in the package, the tests will still pass if they test with the files in the repo instead of the with the installed package. There are other ways to change pytest's behavior but just doing `cd tests` is a simple way to make sure the package is tested.

.github/workflows/python-package.yml
```diff
@@ -31,13 +31,13 @@ jobs:
...
-        cd tests
-        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
+        python -m pytest --cov --cov-report xml --cov-report term
     - name: Upload coverage reports to Codecov
-      uses: codecov/codecov-action@v4.0.1
+      uses: codecov/codecov-action@v4.6.0
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        slug: lmfit/uncertainties
+        file: ./coverage.xml
```
wshanks, line 38: Could we add a `name` entry here as well? If we set the name based on the Python version and OS, would that let us tell which report was from which run? Right now all the reports for PR have the same name on the Codecov page so it is hard to tell which is which. This mainly matters for distinguishing the no-numpy run from everything else.

wshanks (commented):
Although all the checks are green, it seems like something is still wrong with Codecov. When I look at the Upload coverage step of 3.8 ubuntu-latest, I see:

> error - 2024-11-14 13:33:38,751 -- Upload failed: [{"input":{"branch":"jagerber48:feature/code_cov_informational","commit_id":"4026faaab4ee6d4c6163e872defee173ec28988d","repo_id":18933979,"repo_name":"uncertainties"},"loc":["id"],"msg":"Field required","type":"missing","url":"https://errors.pydantic.dev/2.4/v/missing"}]

Looking at the Codecov page, it seems to have only one report instead of one per job in the matrix. I am not sure what report that is. It seems like all uploads failed on the latest run.

jagerber48:
> Although all the checks are green, it seems like something is still wrong with Codecov. When I look at the Upload coverage step of 3.8 ubuntu-latest, I see:
> 
> > error - 2024-11-14 13:33:38,751 -- Upload failed: [{"input":{"branch":"jagerber48:feature/code_cov_informational","commit_id":"4026faaab4ee6d4c6163e872defee173ec28988d","repo_id":18933979,"repo_name":"uncertainties"},"loc":["id"],"msg":"Field required","type":"missing","url":"https://errors.pydantic.dev/2.4/v/missing"}]
> 
> Looking at the Codecov page, it seems to have only one report instead of one per job in the matrix. I am not sure what report that is. It seems like all uploads failed on the latest run.

Yes, see the issue I linked to in codecov feedback. I'm not sure what's going on here. As I said in my other comment, this error goes away if I include the token explicitly, but even though the github action job succeeds, the coverage reports don't appear on codecov.io.

.github/workflows/python-package.yml
jagerber48: I need to understand this more. I think I would prefer to configure the `pytest` runs more explicitly rather than relying on monkeying around with `cd` and the implicit "path".

.github/workflows/python-package.yml
jagerber48: I think `tag`s are what we're looking for here. Yes, we can do that. It's been my plan to add tags, but I wanted to get the more basic configuration working. I doubt adding tags will fix the issues we're seeing, but maybe I'm wrong. Maybe the tags will help.

.github/workflows/python-package.yml
wshanks: That is fine but we can do that in a separate PR? For a long time, the standard way to do this was to use a `src/` directory at the top level of the repo with the package inside of it. More recently pytest added some alternative ways of importing test modules instead of putting the working directory on the Python path. They come with some caveats though that we would have to assess. They are documented here:

https://docs.pytest.org/en/stable/explanation/pythonpath.html

jagerber48: Yes, happy to discuss/do this in a separate PR. I'll pull that change out of this PR when I get a chance.

jagerber48:
It looks like many of the current issues are stemming from the codecov.io server side as per the linked issue. Once those are resolved I think the main push of this PR --- improving and cleaning up codecov config --- can be resolved.

jagerber48:
Ok, seems like https://github.com/codecov/feedback/issues/577 is resolved and our codecov reports are getting uploaded. Will tie up loose ends on this then it will be ready for review.

However, it is troubling the codecov can have strange issues on their end that bothers our CI. In any case, making the code coverage informational only means that these issues don't need slow down our progress if we're happy to move forward even if there are minor coverage issues.

jagerber48:
I've expanded the scope of this PR to include codecov tags. @wshanks any objections? I can move that change into a new PR if we prefer.

jagerber48:
Some more changes:

- I regenerated the codecov token so that it is not the old token that has been revealed in historic source code
- I pinned the test_without_numpy to python 3.12. This clears an error that was happening during the build due to the matrix of python versions not being valid during the no_numpy build.
- I added a flag to each upload

On a previous commit I would put two flags on each with numpy codecov upload. e.g. `windows` and `3.12`. codecov gives a warning that you should only have one flag per upload. I think the issue is that it is kind of strange to have the "windows" flag show the union of the coverage over all python versions. So I heeded codecov's warning and changed the double tags into compound single flags, like `windows-3.12`. This means there are more flags, but I think it means the reporting of coverage over time will be more accurate. Note however that it should be very very rare for us to see an coverage differences between python versions or operating systems.

jagerber48:
This PR is ready for review.

After this I'll transition back to working on source code instead of CI. However, it is on my radar, with low priority, to make some minor improvements to the CI configuration. Mostly just to make the github actions yaml file even simpler and easier to follow. This would include the `cd tests` change discussed above.

newville (commented):
Great!  Thanks for working this out.  I think this is the right approach. 
+1 on merging.

newville (approved):

PR #271, opened 2024-11-16
Refactor and use Cholesky decomposition for `correlated_values` and `correlated_values_norm`

### LINKED ISSUE #99: Use Cholesky decomposition for correlated_values

jagerber48:
- [x] Closes #99 

Clean up `correlated_values` and `correlated_values_norm`. Make the code easier to follow. Also use the more efficient Cholesky decomposition in cases where the user-provided covariance matrix is positive-definite. If it is strictly positive semi-definite then use the eigenvalue decomposition that was previously used.

Note, previously `correlated_values_norm` (which accepts a normalized correlation matrix) did the numerical lifting and `correlated_values` (which accepts a covariance matrix) called `correlated_values_norm`. This is reversed now. Now `correlated_values` does the numerical work and `correlated_values_norm` calls `correlated_values`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,12 +6,17 @@ Unreleased
 
 Changes
 
+- Use a Cholesky decomposition to calculate values with uncertainty using
+   `correlated_values` and `correlated_values_norm` when the provided covariance matrix
+   or correlation matrix is positive definite. If the provided matrix is strictly
+   positive semi-definite an eigenvalue decomposition is still used. Refactor the code
+   and clean up the documentation for these functions. (#99)
 - Changed how `numpy` is handled as an optional dependency. Previously,
    importing a `numpy`-dependent function, like `correlated_values`,
    without `numpy` installed would result in an `ImportError` at import
    time. Now such a function can be imported but if the user attempts to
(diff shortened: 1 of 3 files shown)


jagerber48:
This PR is ready for review.

jagerber48:
While this PR is ready for review, it's going to conflict badly with #262. Let's hold off on this one until after that one.

PR #275, opened 2024-12-16
Add performance test

### LINKED ISSUE #274: Benchmarking tests
The issue concerns adding benchmarking for the performance optimization that was introduced in [release 3.0.1](https://github.com/lmfit/uncertainties/releases/tag/3.0.1). See especially #30 for a thorough discussion. The short summary is that using a naive linear error propagation algorithm will run the following snippet
```
sum(ufloat(1, 0.1) for _ in range(n)).std_dev
```
in O(N^2) time whereas the lazy weight evaluation algorithm innovated for 3.0.1 runs it in O(N) time. I think this is one of the main technical innovations from this package that separates it from anything anyone would try to "whip up".

PR #262 refactors the code that implements this lazy weight evaluation algorithm. For this reason I think it is necessary that we set up benchmarking tests that we confirm we pass before and after merging #262. In fact, I've re-implemented the error propagation multiple times as I've drafted #262 and repeatedly have *failed* to realized the O(N) execution time. This demonstrates the importance of a benchmarking tests to help us ensure we don't accidently introduce a change that ruins the performance (as I already did accidentally during my drafts with just two lines in the wrong place).

________
I have the following simple benchmarking code
```
import platform
import psutil
import timeit

def get_system_info():
    return {
        'platform': platform.system(),
        'platform-release': platform.release(),
        'platform-version': platform.version(),
        'architecture': platform.machine(),
        'processor': platform.processor(),
        'ram': str(round(psutil.virtual_memory().total / (1024.0 ** 3))) + " GB",
    }



def ufloat_sum_benchmark(num):
    str(sum(ufloat(1, 1) for _ in range(num)))


if __name__ == "__main__":
    for key, value in get_system_info().items():
        print(f'{key:17}: {value}')

    for n in (10, 100, 1000, 10000, 100000):
        print(f'### {n=} ###')
        reps = int(100000/n)
        t = timeit.timeit(lambda: ufloat_sum_benchmark(n), number=reps)
        print(f'    Test duration: {t:.2f} s, Repetitions: {reps}')
        print(f'    Average execution time: {t/reps:.4f} s')

```
On my system on `master` branch I get
```
platform         : Windows
platform-release : 10
platform-version : 10.0.19045
architecture     : AMD64
processor        : Intel64 Family 6 Model 154 Stepping 4, GenuineIntel
ram              : 16 GB
### n=10 ###
    Test duration: 0.90 s, Repetitions: 10000
    Average execution time: 0.0001 s
### n=100 ###
    Test duration: 0.79 s, Repetitions: 1000
    Average execution time: 0.0008 s
### n=1000 ###
    Test duration: 0.95 s, Repetitions: 100
    Average execution time: 0.0095 s
### n=10000 ###
    Test duration: 0.99 s, Repetitions: 10
    Average execution time: 0.0995 s
### n=100000 ###
    Test duration: 0.71 s, Repetitions: 1
    Average execution time: 0.7065 s
```
On `feature/linear_combo_refactor` I get
```
platform         : Windows
platform-release : 10
platform-version : 10.0.19045
architecture     : AMD64
processor        : Intel64 Family 6 Model 154 Stepping 4, GenuineIntel
ram              : 16 GB
### n=10 ###
    Test duration: 2.40 s, Repetitions: 10000
    Average execution time: 0.0002 s
### n=100 ###
    Test duration: 2.12 s, Repetitions: 1000
    Average execution time: 0.0021 s
### n=1000 ###
    Test duration: 1.98 s, Repetitions: 100
    Average execution time: 0.0198 s
### n=10000 ###
    Test duration: 1.86 s, Repetitions: 10
    Average execution time: 0.1860 s
### n=100000 ###
    Test duration: 1.77 s, Repetitions: 1
    Average execution time: 1.7692 s
```
____________
So we see that the new code is still linear in time, but between 2x-3x slower than `master`. My guess is the new code could win back that factor of 2x-3x with some careful profiling but I haven't done that yet.

Anyways, the point of this issue is to help me answer the question: What should a benchmarking regression test look like? Perhaps I can run this code under these 5 or so durations and simply make sure the runtimes don't exceed thresholds that are say.... 3x more than the current `master` branch performance? I could also check that the runtime is O(N) to within a factor of say 0.5 - 2 or 0.25 - 4? I'm not sure the right way to set threshold to make sure (1) we catch regressions but (2) we don't have code unluckily failing because it fails the benchmark. Maybe there's a way to make the benchmarking tests "information only" so they don't cause CI to fail, but they do alert code reviewers? Thoughts? I've also asked [this question](https://softwareengineering.stackexchange.com/questions/456030/how-to-test-for-performance-regression) to get even more help with this since it's new to me.

jagerber48:
- [x] Closes #274 

add a performance benchmark test. This test is important especially to ensure #262 doesn't introduce  a performance regression.

### DIFF
--- a/tests/test_performance.py
+++ b/tests/test_performance.py
@@ -7,39 +7,50 @@ import pytest
 from uncertainties import ufloat
 
 
-def ufloat_sum_benchmark(num):
-    sum(ufloat(1, 0.1) for _ in range(num)).std_dev
-
-
-def time_ufloat_sum_benchmark(num):
-    # T ~ N * 10 us, so if we do 1001`000 / N repetitions the test takes ~ 1 s
-    reps = int(100000 / num)
-    t = timeit.timeit(
-        lambda: ufloat_sum_benchmark(num),
-        number=reps,
-        timer=time.process_time,
-    )
-    return t / reps
-
-
-def test_complexity():
+def repeated_summation(num):
     """
-    Test that the execution time is linear in problem size
+    generate and sum many floats together, then calculate the standard deviation of the
+    output. Under the lazy expansion algorithm, the uncertainty remains non-expanded
+    until a request is made to calculate the standard deviation.
     """
-    result_dict = {}
-    n_list = (10, 100, 1000, 10000, 100000)
+    result = sum(ufloat(1, 0.1) for _ in range(num)).std_dev
+    return result
+
+
+def test_repeated_summation_complexity():
+    """
+    Test that the execution time is linear in summation length
+    """
+    approx_execution_time_per_n = 10e-6  # 10 us
+    target_test_duration = 1  # 1 s
+
+    n_list = [10, 100, 1000, 10000, 100000]
+    t_list = []
     for n in n_list:
-        result_dict[n] = time_ufloat_sum_benchmark(n)
+        """
+        Choose the number of repetitions so that the test takes target_test_duration
+        assuming the timing of a single run is approximately
+        N * approx_execution_time_per_n
+        """
+        # Choose the number of repetitions so that the test
+        single_rep_duration = n * approx_execution_time_per_n
+        num_reps = int(target_test_duration / single_rep_duration)
+
+        t_tot = timeit.timeit(
+            lambda: repeated_summation(n),
+            number=num_reps,
+            timer=time.process_time,
+        )
+        t_single = t_tot / num_reps
+        t_list.append(t_single)
     n0 = n_list[0]
-    t0 = result_dict[n0]
-    for n, t in result_dict.items():
-        if n == n0:
-            continue
+    t0 = t_list[0]
+    for n, t in zip(n_list[1:], t_list[1:]):
         # Check that the plot of t vs n is linear on a log scale to within 10%
         assert 0.9 * log10(n / n0) < log10(t / t0) < 1.1 * log10(n / n0)
 
 
 @pytest.mark.parametrize("num", (10, 100, 1000, 10000, 100000))
 @pytest.mark.benchmark
-def test_speed(num):
-    ufloat_sum_benchmark(num)
+def test_repeated_summation_speed(num):
+    repeated_summation(num)
--- a/tests/test_performance.py
+++ b/tests/test_performance.py
@@ -1,0 +1,43 @@
+import platform
+import timeit
+
+import psutil
+
+from uncertainties import ufloat
+
+
+def print_system_info():
+    sys_info = {
+        "platform": platform.system(),
+        "platform-release": platform.release(),
+        "platform-version": platform.version(),
+        "architecture": platform.machine(),
+        "processor": platform.processor(),
+        "ram": str(round(psutil.virtual_memory().total / (1024.0**3))) + " GB",
+    }
+
+    for key, value in sys_info.items():
+        print(f"{key:17}: {value}")
+
+
+def ufloat_sum_benchmark(num):
+    str(sum(ufloat(1, 1) for _ in range(num)))
+
+
+def test_algorithm_performance():
+    print_system_info()
+    result_dict = {}
+    n_list = (10, 100, 1000, 10000, 100000)
+    for n in n_list:
+        print(f"### {n=} ###")
+        reps = int(100000 / n)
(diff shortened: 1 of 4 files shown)


jagerber48:
For the raw benchmark threshold I set that `str(sum(ufloat(1, 0.1) for _ in range(100000)))` should execute in 0.75 * 4 seconds. This came from this code executing in ~0.75 s on my laptop.

However already this threshold was too stringent for the github action runners. My question: how should we put in this benchmark? Should I just increase the factor to 8? Do something else? I'm not sure a good way to test for performance regression since absolute timing is so system dependent.

andrewgsavage:
there's libraries built spefically for this purpose rather than writing your own tests like that. pint uses [codspeed](https://codspeed.io/)

jagerber48:
I made a complexity only test which runs the benchmark for n = [10, 100, ...] and checks that log10(t/t0) is equal to log10(n/n0) to within 10%. Looking at t vs n on a log plot the data is very linear, so I expect this test to be relatively robust. It should definitely catch regressions if t becomes proportional to n^2 instead of n. The worry is that it will occasionally fail erroneously (false negative).
image
x axis is n, y axis is t in seconds. solid lines are linear interpolation between points (no fit or anything)

jagerber48:
I'm hitting a variety of issues with establishing the connection to codspeed.io, but I'll keep working at it.

newville:
@jagerber48 @andrewgsavage Thanks for this.  I approved the codspeed integration request.    

I also agree with the observation in #274 that the lazy evaluation from @lebigot is very impressive and nicely avoids calculations of uncertainties on intermediate quantities that are never needed in isolation.  Clever! And, yes, I can see that this would be easy to slip up when refactoring.    So, thanks, and sorry I can't be more help at this time.

jagerber48:
@newville thank you for approving the integration request. However I see that further configuration is needed. I got a message that the configuration can only be done by an organization owner (which I am not).

This is the configuration that I think needs to be done
image

I access it from the repositories page on my codecov.io account page, then I click to import a new repository (+) and then configure the lmfit organization. Looks like within the `uncertainties` team in `lmfit` org that @newville and @wshanks are owners.

tests/test_performance.py
```diff
@@ -0,0 +1,40 @@
...
+    n0 = n_list[0]
+    t0 = result_dict[n0]
+    for n, t in result_dict.items():
+        if n == n0:
+            continue
+        # Check that the plot of t vs n is linear on a log scale to within 10%
+        assert 0.9 * log10(n / n0) < log10(t / t0) < 1.1 * log10(n / n0)
+
+
+@pytest.mark.benchmark
```
andrewgsavage, line 38: you can probably parameterise this test with n_list and have it benchmark each num

jagerber48: Nesting the benchmark and parametrize marks works in my local environment, but seems to make the github action hang forever running the test.

jagerber48: Ok got it. The problem is I was mixing my internal benchmarking using `timeit` with what codspeed was trying to do. The tool is very nice.

Will cleanup the tests a little bit now that I know what's going on then get this ready for review.

jagerber48:
This is ready for review. @andrewgsavage you seem like the natural person for this one.

tests/test_performance.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    generate and sum many floats together, then calculate the standard deviation of the
+    output. Under the lazy expansion algorithm, the uncertainty remains non-expanded
+    until a request is made to calculate the standard deviation.
+    """
+    result = sum(ufloat(1, 0.1) for _ in range(num)).std_dev
+    return result
+
+
+def test_repeated_summation_complexity():
```
andrewgsavage, line 20: I've seen a test similar to this that kept randomly failing so I don't know how reliable this test will be. 
I think test_repeated_summation_speed will catch any increases in time so these tests are testing the same thing. I'm not opposed to this test- we can leave it in for now, and if it we find it unreliable we can remove it.

I'll add a link to PR with the graph you plotted as it helps makes sense of this test. It'll be good to see this plotted for your other PR too.

andrewgsavage (approved):

PR #277, opened 2024-12-16
Added ufloat_from_sample function

### LINKED ISSUE #276: new ufloat_from_sample function
Hello, I often use this package to analyse data from experiments, and I usually have a list of measurements of a variable. These measurements are usually normally distributed. So I combine them using the mean as the true value and the sample standard deviation divided by the square root of the number of measurements as the standard deviation of the mean. Something like this:
`value=ufloat(numpy.mean(sample),numpy.std(sample,ddof=1)/numpy.sqrt(len(sample)))`

In an attempt to save myself and hopefully a few others some time, would it be suitable to add a function `uncertainties.unumpy.ufloat_from_sample()`

ps I'm new to contributing to open source so if I've made any faux pas, I'm sorry.

Myles244:
- [ ] Closes #276
- [ ] Updated numpy_guide.rst

I'm unsure how to write or run tests, so if someone could point me in the direction of a good resource, it would be much appreciated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Changes
    time. Now such a function can be imported but if the user attempts to
    execute it, a `NotImplementedError` is raised indicating that the
    function can't be used because `numpy` couldn't be imported.
+- Added unumpy.ufloat_from_sample to create a ufloat from a random sample of a variable.
 
 Fixes:
 
--- a/doc/numpy_guide.rst
+++ b/doc/numpy_guide.rst
@@ -141,6 +141,16 @@ and is not named :func:`acos` like in the :mod:`math` module).
 The definition of the mathematical quantities calculated by these
 functions is available in the documentation for  :mod:`uncertainties.umath`.
 
+.. index:: Numbers with uncertainties from samples
+
+Numbers with uncertainties from samples
+---------------------------------------
+
+Numbers with uncertainties can be found from samples of numbers 
+without uncertainties using :func:`ufloat_from_sample`. This is
+an estimate on the true value of the number and its uncertainty,
+and so it works best for large samples that are normally dristobuted.
+
 .. index::
    pair: testing and operations (in arrays); NaN
 
--- a/uncertainties/unumpy/core.py
+++ b/uncertainties/unumpy/core.py
@@ -119,6 +119,16 @@ def std_devs(arr):
 
 ###############################################################################
 
+def ufloat_from_sample(sample):
+    """
+    Return a ufloat where the nominal value and standard deviation are
+    the mean and uncertainty on the mean of the sample of values.
+    If a the sample is a set of random measurements of a normal distribtuion
+    then the returned value will contain the best estimate of the mean
+    and standard deviation of the distribution.
+    """
+    return uncert_core.ufloat(numpy.mean(sample),numpy.std(sample,ddof=1)/numpy.sqrt(len(sample)))
+
 
 def derivative(u, var):
     """


andrewgsavage:
uncertainties uses pytest. 
```
pip install ".[test]"      # to enable running the tests
pytest .
```

will run the tests

see https://docs.pytest.org/en/stable/getting-started.html for a resource 

you'll want to add a test/tests similar to 
https://github.com/lmfit/uncertainties/blob/969324dcbe01dc933888c2540f3b097c99b350f6/tests/test_unumpy.py#L277
testing some example calculations, using `assert result_of_your_function == expected_result`

Myles244:
Thank you.
 I have now completed the following:
- [x] checked that it still passes all tests.
- [x] written an appropriate test for this new function.

jagerber48:
See my comments at #276. Not sure if the discussion should continue here or there. I'm initially wary about adding this in. I see the utility, but I wonder if `uncertainties` wants to get in the business of adding small helper functions like this or not. Also it needs to be made clear (via function names or documentation) that this function should only really be used with data that are known to be normally or approximately normally distributed.

newville (commented):
I like this, and the use of the builtin `statistics` module when Numpy is not available.

Following the precedent from that `statistics` package, I recommend 
`ufloat_from_samples` over `ufloat_from_sample`.
See https://docs.python.org/3/library/statistics.html#statistics.NormalDist.from_samples

I would say this OK in `core.py`.... we could consider moving it somewhere if needed.

uncertainties/core.py
```diff
@@ -1004,6 +1006,76 @@ def ufloat_fromstr(representation, tag=None):
...
+                #if the output is an array of ufloats (duplicate of code from unnumpy.core.uarray to avoid circular import)
+                return numpy.vectorize(
+                    # ! Looking up uncert_core.Variable beforehand through
+                    # '_Variable = uncert_core.Variable' does not result in a
+                    # significant speed up:
+                    lambda v, s: Variable(v, s),
+                    otypes=[object],
+                    )(mean_value, error_on_mean)
+    else:
+        msg={
```
newville, line 1073: Maybe just a f-string:
```
     msg = f"method='{method}' is not implemented"
```

uncertainties/core.py
```diff
@@ -1004,6 +1006,76 @@ def ufloat_fromstr(representation, tag=None):
...
+        the sample is the whole flattened array. 
+    
+    '''
+    
+    if method=="gaussian":
+
+        if numpy is None:
+            #if numpy is not present, use pythons statistics functions instead
+            mean_value=stats_mean(sample)
+            error_on_mean=stats_stdev(sample)/sqrt(len(sample)-1)
```
newville, line 1042: I believe the `statistics.stdev` already divides by `sqrt(len(samples)-1)`, and it should not be done again here.

But also, I think we will need to discuss/document this, and perhaps have an option for using the "-1" ("Bessel's -1"), perhaps adding a `ddof` argument (default=0).   I think the default should be "match numpy.std()", which divides variance by `sqrt(len(samples))`.

uncertainties/core.py
```diff
@@ -1004,6 +1006,76 @@ def ufloat_fromstr(representation, tag=None):
...
+            
+            #the size of each sample being collected
+            sample_size=0
+
+            if axis == None: 
+                sample_size=numpy.size(sample)
+            else:
+                sample_size=numpy.shape(sample)[axis]
+
+            error_on_mean=numpy.std(sample, ddof=1, axis=axis)/numpy.sqrt(sample_size)
```
newville, line 1058: I think the plain `ndarray.std()` should be used by default, and `ddof` should default to 1, and that dividing by the sample size here is not correct.

Myles244: Sorry, I don't understand what you mean. The error on the mean is the variance of the sample divided by the square root of the sample size.

uncertainties/core.py
Myles244: Yes, you are right, i mistakenly assumed that `statistics.stdev` behaved the same as `numpt.std` does without any optional arguments. 

Adding `ddof=0` as an optional parameter should be simple. For the non-numpy code I can use `statistics.pstdev` and then include "-ddof" in the denominator.

uncertainties/core.py
newville: Sorry, I meant ddof should default to 0, as in `numpy.std()`. 

Dividing by sqrt(len(samples)) gives the standard deviation of the mean, not the standard deviation of the value.  
The standard deviation of the mean tells you how much you expect the mean value to change if you select a different set of samples/measurements from a larger pool of values.

For propagating uncertainties, you (usually, normally?) want the standard deviation of the value itself.    

I think two good rule-of-thumbs are "what does Wikipedia say?" and "what does numpy do?".   
Taking the basic example at https://en.wikipedia.org/wiki/Standard_deviation#Basic_examples:
```
>>> import numpy as np
>>> samples = [2, 4, 4, 4, 5, 5, 7, 9]
>>> print(np.mean(samples), np.std(samples))
5.0 2.0
>>> print(np.mean(samples), np.std(samples), ddof=1)
5.0 2.13808993529939
```

which agrees with Wikipedia.    I would be reluctant to have

```
>>> ufloat_from_samples([2, 4, 4, 4, 5, 5, 7, 9])
```
return anything except `ufloat(5.0, 2.0)` - that would require extra explanation and documentation.

An option to divide by `sqrt(len(samples)-ddof)` would be OK.

Myles244: Thank you, I understand now.

I originally intended the function to return the mean and standard error on the mean. This Is because if I am measuring a constant value where my measurement apparatus have some random error, then my measurements are a sample from a normal distribution. Where the best estimate for the value is the mean of the sample and the uncertainty is the uncertainty on the mean.

If you would prefer this not to be the default value, perhaps it could be a method "from measument". I think this would be more clear anyway.

newville: @Myles244 if your samples are from a normal distribution, and you are measuring noisy values from an experimental measurement, then I think you do want to use the mean and the standard deviation (as from numpy.mean() and numpy.std()).   Those give the average (and most likely, for normal distributions) and the appropriate measure of the variation observed.

The standard deviation of the mean is a useful quantity, but I think it is not what you want here.  Let's take mass of Christmas trees (and just to keep things jolly, let's pretend those have Gaussian distributions), and the total sample size in the survey is 1 million trees.  A random sample of 100 would be small, but give a mean and standard deviation, and not be terrible.  If you select a different 100 samples, you would have a mean and standard deviation that was pretty consistent with the first sample of 100. If you increase the sample size to 1000, you would not expect the mean or the standard deviation to change by much from the sample of 100.  But, if you took a second sample of 1000, you would expect that mean value to be closer to the mean for the first 1000 than was the case for the two samples of 100.

Another classic way to think about it is with test scores across many classrooms: give the same test to all students in 10 classrooms of 25 students each. The standard deviation of the 250 scores tells you the variation between students.  The standard deviation of the mean of each classroom tells you the variation between classrooms.  One is a measure of the students; the other is a measure of the teachers (or a bias in how students are assigned classrooms, so the administration, perhaps) ;).

If you are measuring voltages and see a distribution [2, 4, 4, 4, 5, 5, 7, 9], the standard deviation is 2.  If you repeat the measurements 100 times and the values are [[2, 4, 4, 4, 5, 5, 7, 9]*100], the standard deviation is still 2.  

OTOH, if those are 100 different samples of a set of 8 different, possibly related, voltages, then the standard deviation of the mean of those 100 observations of 8 voltages is now much smaller than 2 -- the mean value is very consistent (across those 100 "classrooms").   But to know the difference, I think you would need to provide the 100 mean and standard deviations.

I think the most natural and common use case would be to have a bunch of samples of a single quantity, in which case the standard deviation is wanted.  An option to divide by `sqrt(len(samples)-ddof)` would be OK.

jagerber48: Disclaimer: My main preference is to not have this function added to uncertainties. In my last message I said I was -0. I'm -1 now. What follows will probably be more evidence why I think that is so.

I am with Myles244. I would prefer this function to return a `ufloat` constructed from the mean of the samples and the standard error on the mean (that is `std` divided by the `sqrt` of the number of samples). The samples are drawn from a normal distribution with mean `mu` and variance `sigma^2`. In my measurements I am typically interested in `mu`, not `sigma`. I estimate the distribution mean `mu` with `m` equal to the sample mean. I estimate the distribution variance with `s^2`, the sample variance. But the uncertainty on my estimate `m` of the distribution mean is given by `s/sqrt(n)`, the standard error on the mean. This is the thing which gets smaller as I take more data and this is what I am usually interested in. That is, if I take 4x as much data I can improve my uncertainty on my estimate of the mean by 2x.

So, for my purposes, I prefer something that returns the standard error on the mean rather than the sample standard deviation.

However, other use cases may expect something that returns a sample standard deviation rather than the standard error on the mean! All of this to say, I think this helper function is more domain specific than initially assumed and it probably shouldn't be in `uncertainties`. Just let users do it themselves so they can pick if they want to calculate using unbiased estimators or if they want sample standard deviation or standard error on the mean. We don't need to build all that complexity in when we'll probably get wrong for some folks anyways.

Myles244: @jagerber48 I think I have to agree with you; clearly, the function is ambiguous, and perhaps it would be better not to include it.

Thanks, anyway.

jagerber48: I think you can leave it open for now so we can continue to consider it for at least a few days. While I like the idea of having this as a constructor on `UFloat` I think it is too specific and there are too many different things people don't want. I'm more sympathetic to the idea of a module with a variety of useful functions like this one. But there they could be named and documented more specifically e.g. `ufloat_from_sample_sem`, `ufloat_from_sample_std` etc.

We could include e.g. those two constructor functions on `UFloat`. But in any case, my opinion is that this "utility" work should be held off until we are able to complete some of the "core" maintenance targets we have on `uncertainties` (re-architect the core classes, rework `numpy` integration, eliminate reliance on confusing monkey patching, etc.). In the lifetime of `uncertainties` this new group of maintainers is very new, so I think it makes sense to make changes, especially API changes, slowly as we learn.

andrewgsavage: yes please leave this open

the wiki page mentions the ambiguity:
> When only a [sample](https://en.wikipedia.org/wiki/Statistical_sample) of data from a population is available, the term standard deviation of the sample or sample standard deviation can refer to either the above-mentioned quantity as applied to those data, or to a modified quantity that is an unbiased estimate of the population standard deviation (the standard deviation of the entire population).

so I agree with having explict names like `ufloat_from_sample_sem`, `ufloat_from_sample_std` or a required keyword

jagerber48: @andrewgsavage to be clear, that snippet is from the wikipedia page on [standard deviation](https://en.wikipedia.org/wiki/Standard_deviation) and is referring to the distinction between the sample standard deviation and an unbiased estimator for the population/distribution standard deviation.

Neither of these are the standard error on the mean, which is the uncertainty of an unbiased estimator of the population mean calculated using the sample mean.

So there are three possible options for what the `std_dev` attribute on the returned `UFloat` should be:
- The sample standard deviation
- An unbiased estimator for the population/distribution standard deviation calculated using `ddof != 0`
- The standard error on the mean

I agree that these could be selected by using different functions or a required keyword argument.
____________
If we want this stuff to be in `uncertainties` then here's my proposal:

- We make a new module named `utils.py` (I don't think it should be named `utils.py`, we should try to come up with something better, just using that as a placeholder)
- This module has one function called `ufloat_from_sample_std_dev` which accepts a `ddof` argument, matching the `numpy` API.
- This module has another function called `ufloat_from_sample_std_err`. I don't think any other arguments are necessary other than the samples.

I express a preference for following `numpy` conventions rather than the built-in `statistics` modules conventions. This is because `numpy` is better documented than the `statistics` module. It looks like the `statistics` module is doing some guessing under-the-hood about what `ddof` should equal but it's not really explaining it.

These functions should all work as expected when `numpy` is unavailable. In fact, at first pass, maybe they should be implemented assuming `numpy` is not available. Then we can do benchmarking to see if `numpy` makes a measurable performance difference. If so, we can configure the code to use `numpy` if it is available.

I do NOT think this function should support numpy arrays, at least at first pass. We are still sorting out interoperability with numpy. I think there's a decent chance we'll do something the wrong way and have to redo it if we try to include numpy support for these helper functions. In other words: I could tolerate including helper functions to create `UFloat`, but not helper functions to create arrays of `UFloat` or `UArray` or something.

________________
Other functions can be moved into `utils.py` (under whatever name it gets).

- The functions to create sequences of `UFloat` from nominal values and a covariance or correlation matrix
- The functions to calculate the covariance or correlation matrix from a sequence of `UFloat`
- The `ufloat_from_sample_std_dev` function
- The `ufloat_from_sample_std_err` function
- The uncertainty-weighted mean calculation in #265

jagerber48: Note that one downside of this proposal is that it misses the "slick" opportunity to include e.g. `UFloat.from_sample_std_dev`. But this is ok. At this time, we don't construct `UFloat`s with the class constructor directly anyways, it is done with the `ufloat()` helper function. In the future we can consider moving away from `ufloat` and make `UFloat` the direct constructor. At that time we could consider adding helper methods onto the class for alternative construction methods. But right now, since construction is done through a function, I think it makes sense for alternative construction to also be done through a function.

newville:
@jagerber48 @Myles244 @andrewgsavage I have to admit that, even after re-visiting these comments, I am extremely uncomfortable using the standard deviation of the mean here.  

If there is a function to convert a set/list/ndarray of values to a ufloat and no other information is provided, I think that it must use the mean and standard deviation of the values supplied.  I would need a lot of convincing if  `ufloat_from_samples([2, 4, 4, 4, 5, 5, 7, 9])` does not give `ufloat(5.0, 2.0)`.

I am OK with having a `ddof` argument with a default value of 0.  The function should be close to: 
```python
def ufloat_from_samples(vals, ddof=0):
    if np:
        avals = np.array(vals)
        return ufloat(avals.mean(), avals.std(ddof=ddof))
    else:
        std = statistics.pstdev(vals) if ddof == 0 else statistics.stdev(vals) # only allows ddof=0,1
        return ufloat(statistics.mean(vals), std)
```
  
A ufloat represents a mean or most likely value from a distribution and the standard deviation of that distribution so that uncertainties (of the distribution!) can be propagated   Indeed, the docstrings and the docs for this code call the second quantity of a ufloat "the standard deviation" throughout, and the variable is called `std_dev`, an abbreviation of "standard deviation".   

I do not see a need for a ufloat constructed from the standard error or the standard deviation from the mean.  While the standard deviation of the mean is a useful statistic, it is not used for error propagation.

jagerber48:
We're all in agreement now that it would not be appropriate for `ufloat_from_samples([2, 4, 4, 4, 5, 5, 7, 9])` to return the standard error on the mean. We all agree that if users want the standard error on the mean they either need to use a different, more specifically named function, or they need to pass an obviously named argument. So I don't think you need to worry that `ufloat_from_samples`, without further qualification, will return the standard error on the mean.

PR #289, opened 2025-01-30
unumpy array creation from string representation

jnahlers:
`uncertainties` has the very useful function `ufloat_fromstr` to create `ufloat` objects from various string representations. I thought that it would be a good symmetry to also include a `uarray_fromstr` function to turn string representations of *arrays* of `ufloat` string representations into `uarrays`. In particular, to re-create `uarray` objects from the results of `str(my_uarray)` (independent and  uncorrelated, of course).

```python
>>> import numpy as np
>>> from uncertainties import ufloat, unumpy as unp
>>> my_uarray = np.array([ufloat(2,1), ufloat(1.1,0.1)])
>>> my_str_rep = str(my_uarray)
>>> my_str_rep
'[2.0+/-1.0 1.1+/-0.1]'
>>> my_new_uarray = unp.uarray_fromstr(my_str_rep)
>>> unp.nominal_values(my_uarray) == unp.nominal_values(my_new_uarray)
array([ True,  True])
>>> unp.std_devs(my_uarray) == unp.std_devs(my_new_uarray)
array([ True,  True])
```

Implementing this has a few issues. Firstly, the string representations of `numpy` arrays use spaces to separate values. Therefore, string representations of ufloats that contain spaces (as supported by `ufloat_fromstr`) cannot (straightforwardly) be supported by `uarray_fromstr`.

`uarrays`, just like `numpy` arrays, can be multi-dimensional. In this case, the string representation will contain nested square brackets and newline characters. I have implemented `uarray_fromstr` to support this using [a slightly hacky workaround](https://stackoverflow.com/a/38887164) that relies on [`ast.literal_eval`](https://docs.python.org/3/library/ast.html#ast.literal_eval). While this is much safer than `eval`, there are still [some important security considerations](https://stackoverflow.com/a/7689085).

- [X] The change is fully covered by automated unit tests
- [X] Documented in docs/ as appropriate
- [X] Added an entry to the CHANGES file

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@ Adds:
    O(N), rather than O(N^2), scaling complexity for operations involving many numbers
    with uncertainty. Established connectivity with `codspeed.io<codspeed.io>`_ to track
    benchmarking results. (#274)
+- Added a function `unumpy.uarray_fromstr` to create unumpy arrays from string
+   representations.
 
 
 Fixes:
--- a/doc/numpy_guide.rst
+++ b/doc/numpy_guide.rst
@@ -74,6 +74,13 @@ through NumPy, thanks to NumPy's support of arrays of arbitrary objects:
 
 >>> arr = numpy.array([ufloat(1, 0.1), ufloat(2, 0.002)])
 
+uarray Variables can also be created from a string representation. This follows the
+numpy convention, with square brackets and spaces separating the elements. A number of
+ufloat string representations are supported, as long as they do not contain any
+whitespace:
+
+>>> arr = unumpy.uarray_fromstr("[1.0+/-0.01 2.0+/-0.002]")
+
 .. index::
    single: matrices; creation and manipulation
    single: creation; matrices
--- a/doc/numpy_guide.rst
+++ b/doc/numpy_guide.rst
@@ -80,6 +80,12 @@ ufloat string representations are supported, as long as they do not contain any
 whitespace:
 
 >>> arr = unumpy.uarray_fromstr("[1.0+/-0.01 2.0+/-0.002]")
+>>> arr = unumpy.uarray_fromstr("[1.0(1) 2.0(2)]")
+>>> arr = unumpy.uarray_fromstr("[1.0+/-0.01 2.0(2) (6.283±0.002)E+02]")
+
+Creation from string representations of higher-dimensional arrays is also supported:
+
+>>> arr = unumpy.uarray_fromstr("[[1+/-1 2+/-2]\n [3+/-3 4+/-4]]")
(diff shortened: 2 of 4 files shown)


newville:
@jnahlers Thanks.  I have to admit that I am not an enormous fan of either `ufloat_fromstr` or of numpy arrays of ufloat objects (as opposed to having an object that contained an array of floats and an array of uncertainties).  

I think the challenges and messy bits of `ufloat_fromstr` are compounded for arrays.   

But also:  Is there a problem that this is solving? Like, who has string representations of arrays of ufloats that are stored in a way that needs to be parsed?   That is, as opposed to a file with tables with columns for nominal values and standard errors?

jnahlers:
> I have to admit that I am not an enormous fan of either `ufloat_fromstr` or of numpy arrays of ufloat objects (as opposed to having an object that contained an array of floats and an array of uncertainties).

The beauty of the `unumpy` module (and indeed, the uncertainties package as a whole) is that it enables easy uncertainty propagation (including remembering correlations) through lengthy calculations.  

> I think the challenges and messy bits of `ufloat_fromstr` are compounded for arrays.

Agreed! Certainly `ufloat_fromstr` is messy, and I am very grateful for the work that was put into implementing it. I think I found a fairly clean and straightforward extension to arrays, but of course that could quickly change if issues are found and things get more complicated.

> But also: Is there a problem that this is solving? Like, who has string representations of arrays of ufloats that are stored in a way that needs to be parsed? That is, as opposed to a file with tables with columns for nominal values and standard errors?

Anyone who has to deal will `unumpy` arrays that were dumped to text. I encountered it in the context of a large `pandas` dataframe that included unumpy arrays, which was saved using the dataframe method `to_csv()`. Of course, this is far from ideal, but I think its inevitable that people will dump objects to text and so a parser (which is essentially what this is) can be quite convenient.

andrewgsavage:
It may be more appropriate for your problem to add writing/parsing support to uncertainties-pandas
https://github.com/andrewgsavage/uncertainties-pandas/blob/main/notebooks/tutorial.ipynb

jnahlers:
> It may be more appropriate for your problem to add writing/parsing support to uncertainties-pandas https://github.com/andrewgsavage/uncertainties-pandas/blob/main/notebooks/tutorial.ipynb

Interesting library! I am not working with `pandas` myself, I am just parsing a text output from `pandas`.

andrewgsavage:
my thinking is that whoever is writing the text output from pandas could
use uncertainties-pandas to write the an output file, then you can use
uncertainties-pandas to read the outfile in a DataFrame, then use
.to_numpy() to get the array of uncertainties you want.

PR #290, opened 2025-03-06
Fix small float formatting

### LINKED ISSUE #135: ZeroDivisionError when getting the __repr__ if nominal value is too small
Perhaps this is a very obscure bug but it does exist.

When running the following block of code:

```python
n, s = 1.1E-323, 0
v = uncertainties.core.Variable(n, s)
print(v)
```
It will throw a ZeroDivisionError at the print line.

I understand this is because, when converting the variable into a string using `v.__repr__()` it will attempt to factorize out the 1E-323 from both the v.n and the v.s.

It seems that python would only complain about division by zero when n is between [2.5E324, 12.5E324), making it a very specific and obscure problem, which is perhaps why I'm the first person to file an issue for it
I was working with a large enough dataset to hit upon this weird bug.

I know a few ways to workaround this problem on my own (choose all values <12.5E-324 and then set them to zero); but do you think there's a way to write the `__repr__` in such a way to avoid future occurrences of this?

Side note:
Interestingly while testing the behaviour, I found out that python rounds all numbers above 2.5E-324 to the nearest multiple of 5E-324. I guess this is the resolution of IEEE-754 numbers?

OceanNuclear:
Fixed the edge case of having super-small (but non-zero) AffineFunc expressed as str.

- [x] Closes #135 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,8 @@ Fixes:
 - adjust `codecov.io` configuration so that minor code coverage changes will not result
    in indications that tests are failing. Rather code coverage reports will be purely
    informational for code reviewers. Also fix other minor configuration issues. (#270)
+- fixes to the str representation which occassionally errors when AffineScalarFunc is
+  too small (#135)
 
 3.2.2   2024-July-08
 -----------------------
--- a/tests/test_formatting.py
+++ b/tests/test_formatting.py
@@ -26,6 +26,18 @@ def test_PDG_precision():
     for std_dev, result in tests.items():
         assert formatting.PDG_precision(std_dev) == result
 
+def test_small_float():
+    """
+    Make sure that very small floats do not error, even though printing as str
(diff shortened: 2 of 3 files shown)


newville (approved):
Thanks @OceanNuclear, I think this is all okay.  
I will also say that UFloats with a stderr of zero (or less that 2e-16*nominal_value)  basically make no sense and will likely cause problems.   A value with a (finite) uncertainty can be represented as a UFloat.  A value with an infinitesimal uncertainty is a Float.   I find it odd how much attention UFloats with stderr=0 get. 
I will also say that all of the formatting code in `uncertainties` is a distraction from the main point of the library.   

And yet, I'll go with a bit of a shrug and hearty thanks @OceanNuclear for the fix to this very warty part of the code.

newville:
@OceanNuclear merging... thanks!

OceanNuclear:
Glad to be of help!

Issue #283, opened 2024-12-31
Problem with special casing `std_dev = 0`


jagerber48:
This is a subtle problem that has been an Achilles heel of #262. On master branch we have

```
from uncertainties import ufloat

zero = ufloat(0, 0)
p = ufloat(0.3, 0.01)
assert (zero ** p).n == 0
assert (zero ** p).s == 0
```

Let `f=x**y`. In this case with `x==0` and non-integer `y` we would expect `df/dx` to be `nan`. Indeed this is what is returned by `pow_deriv_0` in `ops.py` (the function responsible for this math) https://github.com/lmfit/uncertainties/blob/master/uncertainties/ops.py#L137. 
We would expect `df/dy` to be 0 in this case and this is what is returned by `pow_deriv_1`. So I would expect the uncertainty to be

```
sqrt((df/dy * 0.01)**2 + (df/dx * 0)**2)
```
But since `df/dx` is `nan` the whole equation will be converted to `nan`. So, I would expect `(zero**p).s` to have a `nan` uncertainty. However, we can see that it does not. The reason it does not is because of this check in `AffineScalarFunc.error_components` at line 509

https://github.com/lmfit/uncertainties/blob/61f688f920b34d06da521a5fb8b013043e2e2fe0/uncertainties/core.py#L500-L516

Basically if any variable has zero std dev then it is excluded from the uncertainty calculation. My read of this is that `uncertainties` has attempted to treat `UFloat` with zero std dev as a float.

This has been... tricky.. to implement in #262. I was able to get it at one point, but it required a lot of special casing in various places. However, thinking about it so much has led me to discover what I would call a bug.
________________
```
from math import isnan

from uncertainties import ufloat

zero = ufloat(0, 0)
p = ufloat(0.3, 0.01)
assert (zero ** p).n == 0
assert (zero ** p).s == 0

x = ufloat(1, 0.1)
x_zero = x-x
assert x_zero == zero
assert (x_zero ** p).n == 0
assert isnan((x_zero ** p).s)
```
In the latter case we see `x_zero ** p` has `nan` std_dev despite `x_zero == zero`. The problem is the catch on variables with zero std_dev only works on `Variable`. But `x_zero`, is not a `Variable`, it is an `AffineScalarFunc`. We also can't compute the std_dev of `AffineScalarFunc` mid-computation because this is the costly step that would spoil the O(N) lazy linear combination expansion algorithm.
________________
[...]

PR #292, opened 2025-03-12
Clean up `pow` tests

### LINKED ISSUE #291: Clean up power functions tests
The tests testing the action of the `pow` and `umath_core.pow` operators on `UFloat` instances are hard to read and poorly named.  The tests should be refactored to improve readability. 

This issue is related to issues 

- #273 because it involves cleaning up tests
- #283 because these are some of the main tests that involve equality comparison between `UFloat` and `float` instances
- #262 because these tests, and especially their relationship to #283 have given me a lot of headaches while working on #262.

Recommendations for improvements:
- Use `pytest.mark.parametrize`.
- Try to unify the tests in `test_uncertainties` and `test_umath`

jagerber48:
- [x] Closes #291

Summary of goals:

- Changes all power tests to use `pytest.mark.parametrize`. Hopefully this makes the tests more readable. 
- Combines the tests for `pow` and `umath_core.pow` in a `test_power.py` module to get some code re-use and to combine similar tests rather than spread them across modules
- Remove `pow` test code from `helpers.py` module.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -5,6 +5,13 @@ import uncertainties.core as uncert_core
 from uncertainties.core import ufloat, AffineScalarFunc
 
 
+def nan_close(first, second):
+    if isnan(first):
+        return isnan(second)
+    else:
+        return isclose(first, second)
+
+
 zero = ufloat(0, 0.1)
 zero2 = ufloat(0, 0.1)
 one = ufloat(1, 0.1)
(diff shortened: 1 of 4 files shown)


newville:
@jagerber48 Thanks - this looks like an improvement in readability to me!  I do worry a bit about a test with 
```integer = ufloat(-3, 0)``` 

It looks like that is only used to test that using that as an exponent yields a stderr of NaN.    That's OK.
I would also be OK with deliberately not testing that:  the value `ufloat(3, 0)` makes no sense,  so we should feel obligated to promise a sensible result when using it.

+1 on merging.

newville (approved):
Looks good to me.  Thanks!

jagerber48:
I'll double check that all cases are copied over correctly for now and then merge it in.

> I would also be OK with deliberately not testing that: the value ufloat(3, 0) makes no sense, so we should feel obligated to promise a sensible result when using it.

Definitely interested in this idea since it's tests like these that have been giving me a lot of trouble on #262 and caused to me raise #283. We can discuss elsewhere. I'll keep this PR as a simple 1:1 test refactor.

PR #294, opened 2025-03-14
Refactor power derivative functions

### LINKED ISSUE #293: Clean up power derivative calculations
`uncertainties` calculates the `x` and `y` derivatives of `pow(x, y) = x**y` using the `pow_deriv_0` and `pow_deriv_1` local functions defined in `get_ops_with_reflection` in `ops.py`.

https://github.com/lmfit/uncertainties/blob/19865cc0e87558d1d105c9d674ebace501e41395/uncertainties/ops.py#L131-L143

My issues with `pow_deriv_0`:

- It returns complex results for `x<0` with non-integer `y` even though `uncertainties` only supports real values. I would prefer it to return `nan` or raise a `ValueError` (weak preference for returning `nan`).
- For `x=0`, `y<0` and `y` an integer it raises a `ZeroDivisionError`, but for `x=0`, `y<0` and `y` a non-integer it returns `nan`. I think for `x=0` and `y<0` this function should always return `nan` or should always raise an exception (weak preference for returning `nan`).

My issues with `pow_deriv_1`:

- It raises a `ValueError` for all `x<0`. This is different than `pow_deriv_0` which returns complex results for this case. It should have the same behavior as `pow_deriv_0`, either return `nan` (my weak preference) or raise a `ValueError`.
- For `x=0` and `y<=0` it always raises `ValueError`. This is ok but again it's not consistent with the `pow_deriv_0` behavior. The two should be consistent, my preference would be to return `nan` or raise `ValueError`.

Other issues with both:
 
- But one big issue with both of these is that when the `ValueError` is raised, it is only raised as a consequence of the derivative calculation and is raised by built-in functions. It would be preferable to explicitly slice up the domain and handle all cases explicitly rather than relying on implicit behavior of the functions involved.
- They are both defined locally in the scope of another function `get_ops_with_reflection` in `ops.py`. They should be defined top-level in the module. This will allow direct testing to help clarify these edge cases.

jagerber48:
- [x] Closes #293

`uncertainties` allows calculating `x**y` where one or both of `x` and `y` are `UFloat` objects. For this we need to calculate the `x` and `y` derivatives of `x**y`. This PR refactors the functions that calculate those derivatives to more explicitly and clearly handle non-typical cases. Moving those functions to the module level also allows these functions to be directly tested.


newville (dismissed):
Looks good to me.  Thanks for adding those tests.

uncertainties/ops.py
```diff
@@ -119,29 +141,6 @@ def get_ops_with_reflection():
             eval("lambda y, x: %s" % expr) for expr in reversed(derivatives)
         ]
 
-    # The derivatives of pow() are more complicated:
-
-    # The case x**y is constant one the line x = 0 and in y = 0;
```
andrewgsavage, line 124: it's a shame to lose this comment. is there somewhere it could be moved to? I would say a docstring but that's not useful for `__pow__`.

maybe it could be added to `umath.pow`'s docstring?

jagerber48: Let me take a look. I actually spent a VERY long time puzzling over various special cases and I had some rewrites of this same comment.

Part of the problem is that the old function handled different special cases using different approaches. For example, the old function allowed complex results to be returned but, instead, relied on some other part of the codebase to prevent complex results.

Under this PR the comment is not actually correct anymore but I can come up with replacement text. But I'm curious, do you want a comment for the developer's sake (like the old comment used to be) or for the user's sake? For the latter we can add some description under the `umath.pow` docstring. We can also add documentation to the docs webpage. If we want a comment for the developers sake then I can update the old comments with some notes about the various special cases and how they're handled.

Also happy to do all of the above, though it ends up being a lot of text because there's a few different cases.

andrewgsavage: I think it's most useful for users so a docstring or section of the documentation, either works.

andrewgsavage:
seems good

jagerber48:
Cleaned it up a little more. Think it's ready for approval and merging if others agree.

jagerber48:
Can someone please approve and merge #302 and then merge this PR? @newville @andrewgsavage? After that we are ready to release version `3.2.3` and begin work on the `4.0.0` release.

newville (approved):
@jagerber48 @andrewgsavage  This looks good to me...  Any objections to merging?

andrewgsavage:
go for it

PR #297, opened 2025-03-19
Remove compare derivatives test function

### LINKED ISSUE #296: The `compare_derivatives` test function is too complicated
`uncertainties` calculates derivatives of various math functions, either instance methods on `UFloat` or module functions in `umath`, using hard-coded analytic formulas or using numerical differentiation. As far as I can tell the `compare_derivatives` test function is used to compare that the method used to calculate derivatives for `UFloat` methods or `umath` functions gives a result close to the corresponding numerical derivative. However, there are a few issues with the `compare_derivatives` test function.

- It is very complicated It is hard to follow the test cases through this function. There are so many nested conditionals and locally defined functions. There is also some complicated `try... except...` logic and retrying. 
- The `compare_derivatives` function is looped over, but the corresponding tests don't use `pytest` parametrization, so it can be hard to track which tests failed or why.
- The `compare_derivatives` function uses random sample data generated at test time for testing. This is a bad practice. What if the tests happen to rarely fail for some random input? We might see the test fail once spuriously but then work. Unless we can access the test data for the failing case we won't be able to diagnose or resolve the spurious failure.
- This helper function for `uncertainties` and `umath` tests doesn't really need to be in its own module.

jagerber48:
- [x] Closes #296 (insert issue number)

Refactor the derivatives comparison tests to eliminate the `compare_derivatives` test function in `helpers.py`. Right now the complication of this test function is making addressing the topics in #295 difficult.

This PR only impacts testing code.  No code for actual usage is removed. A `NumericalDerivatives` helper class was removed from `core.py` but this class was only only used for testing and was not documented for end-user usage.

Some `UFloat` methods and `umath` functions which were previously tested for `UFloat`-calculated-derivative compared to numerical derivatives are no longer tested in this way. The methods/functions for which testing was dropped are listed below. These functions have been discussed in #295 where there is at least tacit agreement that `uncertainties` can drop these functions/methods entirely. So dropping their tests should be uncontroversial.

The following `UFloat` methods are no longer tested in this way:

- `__abs__`
- `__trunc__`
- `__floordiv__`
- `__rfloordiv__`
- `__mod__`
- `__rmod__`

The following `umath` functions are no longer tested in this way:

- `copysign`
- `fabs`
- `fmod`
- `ldexp`
- `frac_part_modf`


jagerber48:
Note that, in anticipation of agreement on #295, this change does not have derivative tests for

- `UFloat.__abs__`
- `UFloat.__trunc__`
- `umath.fmod`
- `umath.ceil`
- `umath.floor`
- `umath.isinf`
- `umath.trunc`
- `umath.copysign`
- `umath.fabs`
- `umath.factorial`
- `umath.modf`
- `umath.ldexp`
- `umath.frexp`

I think in some cases this is dropping some testing compared to the base and in some cases these functions weren't tested in this way anyways.

newville (dismissed):
I think import statements on two of tests modules need to be brought to current master  but otherwise, this looks good -- thanks @jagerber48.

jagerber48:
@newville thanks for taking a look. Looks like it's good to go after you merged in `master`. If you don't merge it in by Friday I think I'll give it a review myself if I can and merge it in (since you basically already approved it, it just got dismissed by the merge I think.

newville (approved):
Looks good to me. Thanks.

PR #298, opened 2025-03-24
Add deprecation warnings for `umath` functions and `UFloat` methods

jagerber48:

Add deprecation warnings for various `UFloat` methods and `umath` functions. See the list of deprecated methods/functions in the source code and see also #295.

### DIFF
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -1054,7 +1054,7 @@ def ufloat(nominal_value, std_dev=None, tag=None):
 def deprecation_wrapper(func, msg):
     @functools.wraps(func)
     def wrapped(*args, **kwargs):
-        warn(msg, FutureWarning)
+        warn(msg, FutureWarning, stacklevel=2)
         return func(*args, **kwargs)
 
     return wrapped
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -15,7 +15,9 @@ Main module for the uncertainties package, with internal functions.
 from __future__ import division  # Many analytical derivatives depend on this
 
 from builtins import str, zip, range, object
+import functools
 from math import sqrt, isfinite  # Optimization: no attribute look-up
+from warnings import warn
 
 import copy
 import collections
@@ -1044,3 +1046,34 @@ def ufloat(nominal_value, std_dev=None, tag=None):
     """
 
     return Variable(nominal_value, std_dev, tag=tag)
+
+
+# Deprecated UFloat methods
+
+
+def deprecation_wrapper(func, msg):
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        warn(msg, FutureWarning)
+        return func(*args, **kwargs)
+
+    return wrapped
+
+
+deprecated_methods = [
+    "__floordiv__",
+    "__mod__",
+    "__abs__",
+    "__trunc__",
+]
+
+for method_name in deprecated_methods:
+    message = (
+        f"AffineScalarFunc.{method_name}() is deprecated. It will be removed in a future "
+        f"release."
+    )
+    setattr(
+        AffineScalarFunc,
+        method_name,
+        deprecation_wrapper(getattr(AffineScalarFunc, method_name), message),
+    )
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,9 +38,9 @@ Deprecates:
 - Certain `umath` functions and `AffineScalarFunc`/`UFloat` methods will be removed in
    a future release. A deprecation warning has been added to these functions and
    methods. The following `umath` functions are marked as deprecated: `ceil`,
-   `copysign`, `isinf`, `isnan`, `fabs`, `factorial`, `floor`, `fmod`, `frexp`, `ldexp`,
-   `modf`, `trunc`. The following `AffineScalarFunc`/`UFloat` methods are marked as
-   deprecated: `__floordiv__`, `__mod__`, `__abs__`, `__trunc__`.
+   `copysign`, `fabs`, `factorial`, `floor`, `fmod`, `frexp`, `ldexp`, `modf`, `trunc`.
+   The following `AffineScalarFunc`/`UFloat` methods are marked as deprecated:
+   `__floordiv__`, `__mod__`, `__abs__`, `__trunc__`.
 
 3.2.2   2024-July-08
 -----------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,15 @@ Fixes:
 - fixes to the str representation which occassionally errors when AffineScalarFunc is
   too small (#135)
 
+Deprecates:
+
+- Certain `umath` functions and `AffineScalarFunc`/`UFloat` methods will be removed in
+   a future release. A deprecation warning has been added to these functions and
+   methods. The following `umath` functions are marked as deprecated: `ceil`,
+   `copysign`, `isinf`, `isnan`, `fabs`, `factorial`, `floor`, `fmod`, `frexp`, `ldexp`,
+   `modf`, `trunc`. The following `AffineScalarFunc`/`UFloat` methods are marked as
+   deprecated: `__floordiv__`, `__mod__`, `__abs__`, `__trunc__`.
+
 3.2.2   2024-July-08
 -----------------------
 
--- a/tests/test_umath.py
+++ b/tests/test_umath.py
@@ -1,6 +1,8 @@
+import inspect
 import math
 from math import isnan
 
+import pytest
 
 from uncertainties import ufloat
 import uncertainties.core as uncert_core
@@ -274,3 +276,19 @@ def test_hypot():
     result = umath_core.hypot(x, y)
     assert isnan(result.derivatives[x])
     assert isnan(result.derivatives[y])
+
+
+@pytest.mark.parametrize("function_name", umath_core.deprecated_functions)
+def test_deprecated_function(function_name):
+    num_args = len(inspect.signature(getattr(math, function_name)).parameters)
+    args = [ufloat(1, 0.1)]
+    if num_args == 1:
+        if function_name == "factorial":
+            args[0] = 6
+    else:
+        if function_name == "ldexp":
+            args.append(3)
+        else:
+            args.append(ufloat(-12, 2.4))
+    with pytest.warns(FutureWarning, match="will be removed"):
+        getattr(umath_core, function_name)(*args)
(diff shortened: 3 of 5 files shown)


jagerber48:
@andrewgsavage in the other thread you expressed that you think we should keep `isinf` and `isnan`. Right now they are marked as deprecated, but I'm open to removing them from the deprecation list. The reason I continue to want to remove them is, again, it's not 100% obvious how these should behave on `UFloat`. Right now if `x` is any `UFloat` we always have
```
import math
import umath

assert math.isnan(x) == umath.isnan(x.n)
assert math.isinf(x) == umath.isinf(x.n)
```
But what if the nominal value is finite but the std_dev is infinite? Should those trigger inf/nan flags? I could see use cases where the answer is yes, and I could see some users being surprised. Since it is so simple to do `math.isnan(x.n)` instead of `umath.isnan(x)` I think we should get rid of the `umath` versions and let users do whatever comparison they want to do, i.e., if they want to check if the nominal value is finite or nan they should explicitly spell that out.

uncertainties/core.py
```diff
@@ -1044,3 +1046,34 @@ def ufloat(nominal_value, std_dev=None, tag=None):
...
     return Variable(nominal_value, std_dev, tag=tag)
+
+
+# Deprecated UFloat methods
+
+
+def deprecation_wrapper(func, msg):
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        warn(msg, FutureWarning)
```
wshanks, line 1057: I think you want stackevel 2 so the warning doesn't blame the wrapper function.

```suggestion
        warn(msg, FutureWarning, stacklevel=2)
```

wshanks:
I think things like `isnan` are the way they are to facilitate using them interchangeably with `float` and `UFloat`. I agree that they are always used with `UFloat` it is just as easy to use `.n` and not have the special function. On the other hand, the documentation is pretty clear that they only work on the nominal value and they don't seem hard to maintain. So I am a bit torn. On the one hand, I like moving away from the complexity of treating `UFloat` and `float` interchangeably. On the other, it would be a breaking change and the current way is not hard to maintain. The documentation does cover [a specific use case for isnan](https://uncertainties.readthedocs.io/en/latest/numpy_guide.html#nan-testing-and-nan-aware-operations).

jagerber48:
@wshanks thanks for the comment. 

Regarding the maintenance burden of `isinf` and `isnan`. Right now there are many different function type and wrappers in `op.py` and `umath_core.py`. In the future if we only generalize functions `foo` that take floats as inputs and return floats as outputs then we can have a single wrapper that allows `foo` to take `UFloat` or `float` as input by looping through the input arguments and applying the uncertainty propagation formula to any `UFloat`s passed in. `isnan` and `isinf` break this paradigm because (1) they don't return `float` and (2) they require the special behavior of picking out the nominal value. So, relative to that vision I have, functions like `isinf` and `isnan` as they behave now will always require some special casing in the implementation, tests, and documentation. For example, right now `isinf` and `isnan` would be the only remaining functions in `locally_cst_funcs` in `umath_core.py`. Deleting `locally_cst_function` entirely would allow us to eliminate a lot of lines of code in these confusing modules.

As for the array example in the docs: In that example `unumpy.isnan` is used to generate a mask from a `unumpy` array. But this can easily be done by the user by taking the nominal value numpy array first then using `numpy.isnan`. And hey, maybe users want to mask out values with `nan` `std_dev` in addition `nan` values so that one `nan` `std_dev` in an array of 10000 `UFloat` doesn't spoil the carefully calculated `std_dev` of the other 9999 `UFloat`s.

wshanks:
In the docs example the key line is

```python
arr[~unumpy.isnan(arr)].mean()
```

Implementing that without a helper function would look like:

```python
arr[~numpy.isnan(unumpy.nominal_values(arr))].mean()
```

I agree that it doesn't save much code. It also might be the case that the uncertainties code is overengineered to avoid repetition with its various wrappers and loops. Unless I am overlooking some edge cases, the implementation of `unumpy.isnan` could be simply

```python
def isnan(arr):
    return numpy.isnan(nominal_values(arr))
```

which is why I had said it didn't seem hard to maintain.

wshanks:
Okay, I think I am okay with deprecating `isnan` but it would be good to hear from @andrewgsavage. For some of these things that don't fit the main path we want to support I would be okay with marking them deprecated and removing any problematic tests that impede refactoring like #262. Otherwise though I don't think there is a need to remove them quickly (at least for things like `isnan` where it seems like it could just be a one line function).

jagerber48:
I hear your sentiment.

> Otherwise though I don't think there is a need to remove them quickly (at least for things like isnan where it seems like it could just be a one line function).

Right now it's not really a one-liner because the inclusion of these functions results in a lot of other infrastructure decorators, conditional branches etc. However, they could be refactored to be one-liners and eliminate all of this infrastructure. That would be an improvement in the code base which is the goal.

I think my problem is I can imagine use cases where a user might want `isnan` or `isinf` to ALSO check the `std_dev`. It feels like an arbitrary decision on the side of `uncertainties` to say whether or not `isnan` and `isinf` look at `std_dev`. The two sides are basically: 

- Yes it's an arbitrary decision, let's remove it and leave it up to the user (breaks backwards compatibility for some users)
- Yes it's an arbitrary historical decision. But it's useful for some users as is. For the sake of not breaking backwards compatibility don't remove it.

My preference is still to remove them for 4.0 when we're making other breaking changes. But if there's strong opinions the other way I would be fine to leave `isnan` and `isinf` and refactor them as described above in another PR.

andrewgsavage:
I haven't seen an issue where users have been confused by isnan or isinf and it is clear in the docs as to what they do. 
I also disagree with it being an arbitrary decision. It's a design decision which, in almost all cases, behaves as the user expects. nan handling is so common that having to do .nominal_values degrades useability somewhat; I think this is worth some maintance burden

I would much rather keep isnan and isinf, and raise a warning when they encounter a ufloat with a nan std_dev.

jagerber48:
@andrewgsavage ok sounds good, I will remove `isnan` and `isinf` from the deprecations in this PR. I'll open a new issue capturing the two improvements we should make to `isnan` and `isnan`, namely:

(1) The logic should be simplified so that we can remove the excess wrapper etc. infrastructure
(2) a warning will be raised in cases in which the `std_dev` is abnormal in a way that might be important to the user.

newville:
@jagerber48 @andrewgsavage FWIW, I think that if I came upon `umath.isnan` (and `isinf`) without carefully reading the docs, I would have expected:
```python
def umath.isnan(uval):
    return math.isnan(uval.n) or math.isnan(uval.s)
```

That would be more like `cmath.isnan`:   `cmath.isnan(2.3 + 1.0j*math.nan)` is True.

But I think the current behavior (`return math.isnan(uval.n)`) is OK.  It's hard to call it the only obvious choice, but it is defensible.

I would be OK with dropping `umath.isnan()` and `umath.isinf()` completely, effectively forcing the user to assert which portions are being tested.  That would need deprecation and documentation and could be a slow process.     

With all of that, I have no strong opinion and defer to the rest of you on this.

jagerber48:
It seems like we have differing opinions here. I agree with Newville that the current behavior is defensible but not the "obvious" right choice. If we were starting over I'd prefer to not have the function.

But anyways, since there is a moderate defensible preference to keep it I'm happy to keep it and work around other ways. I like the idea of warning when users are in ambiguous cases. I also want to keep open the possibility of deprecating in the future. It will also be possible to clean up the code around these functions in the future. I'll open an issue about it.

For now, I think this PR is ready for review. I'm planning to make PRs for the other issues discussed in #295 as we become decided on them.

jagerber48:
Bump. This PR is ready for review.

newville (approved):
Thanks - looks good to me.   Maybe we can come  back to thinking about deprecating `isnan` and `isinf` for version 4.0.   But, let's take these deprecations now.

andrewgsavage:
looks fine

newville:
@andrewgsavage @jagerber48 Thanks - I'll merge (with rebase+merge as it captures the story better).

jagerber48:
@newville why did you rebase + merge rather than just merge? What are the advantages? It added many small commits to the master branch? I would prefer one commit (the merge) to the master branch per pull request. I also don't want to delete the `function_method_deprecation_warnings` branch because then I think those commits will disappear from git as they weren't actually merged into `master`.

newville:
@jagerber48 Oh, sorry... I thought that rebase+merge better captured the history of the changes. and help explain the process.   One commit per merge that touches many files or represents large changes to a few files is sometimes difficult for tracing problems, say, when using `git bisect`.    FWIW, it's not a strong opinion in favor or against any one approach. I would be reluctant to have a strict policy.    Here, I thought the history was worth preserving.   But, one could rebase that all away, if one wanted.

> I also don't want to delete the function_method_deprecation_warnings branch because then I think those commits will disappear from git as they weren't actually merged into master.

Some branches used during developments will be deleted.   These are especially true when using "Feature branches" as Github encourages.   If you want to keep the history of commits, then merging them into master will do that ;)

jagerber48:
@newville thanks for the explanation. Just trying to understand. I agree with no strong opinion and no strict policy. I'm continuing to get used to the github workflow. I'm used to a small team working in a repo without forks. 

Maybe part of my confusion comes from the fact that it seems like merges from fork PRs look different in the commit history than merges within the same repo. 
In `sciform` (where the main repo is my repo) when I merge a PR it looks like this (as viewed in gitkraken):
image

But in this repo it looks like 
image
You can see that commit at the `master` branch only has one parent even though it was the result of merging in the PR. But I guess the problem is the commit that was "merged in" came from my fork and never existed on the main `lmfit` fork, so this is just how github makes the commit history look.

If I'm gathering it all here then I guess what you did, rebase + merge, is a way to bring the history from the forked repo into the main repo so that the history appears as you say.

Hope I'm not overcomplicating it. I'm going to research it on my own a little more.

jagerber48:
Oh no, I see. The behavior I'm seeing is the "squash merge" behavior. It's not actually a merge, it's more like squash + rebase. It takes all the diffs and puts it as one commit on the `master` branch but it's not a merge commit. The behavior I'm used to is a "normal" merge commit, but that's not allowed on this repo. So yeah, if merge commit isn't allowed then rebase merge, instead of squash merge, is the way to keep the history.

newville:
@jagerber48  Yeah, I think I was unsure about why "rebase and merge" was needed, instead of "just merge".  
I am definitely not claiming "complete mastery of all of git/GitHub".   But "Squash and Merge" sometimes seems to emphasize the "squash" ;).

PR #301, opened 2025-04-12
Fix doctests

### LINKED ISSUE #281: Get doctests to pass
During #279 I discovered that there are doctests in the documentation, however, many are failing. On my tests run 111/204 doctests failed. We should get these doctests to pass. It would also be a good idea to run doctests to run in the CI.

jagerber48:
- [ ] Closes #281 

It will be valuable to have all doctests passing. Especially for the release of version `4.0.0`. Right now I see 111/206 doctests failing. I will slowly work through them but, especially on this PR, help from others would be appreciated.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -33,6 +33,7 @@ jobs:
       run: |
         cd tests
         python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
+        python -m sphinx -b doctest doc/ doc/build/doctest/
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,8 +31,7 @@ jobs:
         python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        cd tests
-        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
+        python -m pytest --cov=uncertainties --cov=tests/ --cov-report=xml --cov-report=term
         python -m sphinx -b doctest doc/ doc/build/doctest/
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,7 +31,7 @@ jobs:
         python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        python -m pytest tests/ doc/ uncertainties/ --cov=uncertainties --cov-report=xml --cov-report=term --doctest-modules
+        python -m pytest tests/ doc/ uncertainties/ --cov=uncertainties --cov-report=xml --cov-report=term --doctest-modules --doctest-glob="*.rst"
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .[arrays,test]
+        python -m pip install .[all]
     - name: Test source code and docs
       run: |
         cd tests
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,8 +31,7 @@ jobs:
         python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        python -m pytest --cov=uncertainties --cov=tests/ uncertainties/ --cov-report=xml --cov-report=term --doctest-modules
-        python -m sphinx -b doctest doc/ doc/build/doctest/
+        python -m pytest tests/ doc/ uncertainties/ --cov=uncertainties --cov-report=xml --cov-report=term --doctest-modules
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,7 +31,7 @@ jobs:
         python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        python -m pytest --cov=uncertainties --cov=tests/ --cov-report=xml --cov-report=term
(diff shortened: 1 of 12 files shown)


jagerber48:
I added `scipy` as a dependency for running tests so that the `uncertainties.wrap` function could be practically tested in the documentation. This is a bit heavy-weight, but it was the easiest way I could think of to get a function that doesn't have a simple analytical form to be tested. It fits the documentation exactly "This is for instance
useful when calling external functions (which are out of the user's
control), including functions written in C or Fortran."

newville (dismissed):
@jagerber48 Thanks!  adding `scipy` as a test dependency seems fine to me.   I could have been convinced to give up doctests. But it is impressive that they pass(!), and with this change probably not too hard to keep them working... so I'm in favor of not dropping them.

jagerber48:
I'm going to continue to work on this PR until all the doctests are passing. Will probably be at least a few days or weeks depending on my flexibility

jagerber48:
Many of the test results for arrays are nasty since they're shown with arbitrary precision. I need to figure out way to truncate the precision that also doesn't "get in the way" of the documentation code.

jagerber48:
TODO: run the doctests in CI

jagerber48:
Ugh, the tests of OS/python version dependent rounding differences. There's also a pair of tests that save and load a file that has OS-dependent differences. Need to deal with these.

.github/workflows/python-package.yml
```diff
@@ -28,11 +28,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .[arrays,test]
+        python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        cd tests
-        python -m pytest --cov=uncertainties --cov=. --cov-report=xml --cov-report=term
+        python -m pytest tests/ doc/ uncertainties/ --cov=uncertainties --cov-report=xml --cov-report=term --doctest-modules --doctest-glob="*.rst"
```
jagerber48, line 34: The directories immediately following `pytest` are what are tested. The directory after `--cov` is the directory whose coverage is reported. `--doctest-modules` runs the tests in docstrings in `.py` files. `--doctest-glob="*.rst"` lets the `.rst` files in `doc/` directory be detected for doctest testing.

doc/numpy_guide.rst
```diff
@@ -180,39 +186,65 @@ values.
...
-  2.0+/-0.002
-
-The file can then be read back by instructing NumPy with :meth:`numpy.loadtxt`,
-but for object arrays, this requires a converter function for each column
-separately.  We can use func:`uncertainties.ufloat_fromstr`, but
-:meth:`numpy.loadtxt` passes bytes to converters, they must first be converted
-into a string.  In addition the number of maximum number of columns must be
-known.  An example of using all of this to unpack the data saved with
-:meth:`numpy.savetxt` would be:
+Number with uncertainties can easy be cast to strings and back. This means that arrays
```
jagerber48, line 189: This entire section was changed. The old test was failing in a python/numpy-version dependent way on the `np.loadtxt` call. I couldn't figure out what was happening. This old version of the section ended up being more a tutorial on `np.savetxt` and `np.loadtxt` so I just wrote a different serialization/deserialization example. I think the emphasis on pickling should be removed from the documentation. Pickling may change some in `4.0.0` anyways.

uncertainties/unumpy/core.py
```diff
@@ -500,21 +500,15 @@ def inv_with_derivatives(arr, input_type, derivatives):
...
 
     The result is a unumpy.matrix if numpy.linalg.pinv would return a
     matrix for the array of nominal values.
 
     Analytical formulas are used.
-
-    Original documentation:
-    %s
     """
-    % numpy.linalg.inv.__doc__
```
jagerber48, line 516: The numpy docstring was failing the doctest.. it's also strange to be inserting 3rd party docstrings like this

uncertainties/unumpy/core.py
```diff
@@ -598,11 +592,7 @@ def pinv(array_like, rcond=pinv_default):
     matrix for the array of nominal values.
 
     Analytical formulas are used.
-
-    Original documentation:
-    %s
     """
-    % numpy.linalg.pinv.__doc__
```
jagerber48, line 605: The numpy docstring was failing the doctest.. it's also strange to be inserting 3rd party docstrings like this

jagerber48:
Ok, doctests and docstrings are now being tested in CI and they are all passing.

The one remaining thing I would do is make sure the docs still "look ok". For example, making it so that not too many decimal places are shown in arrays. Basically to get the docs to this point I just worked as fast as possible just making the tests pass by copy and pasting their current results. 

Others can weigh in on if that is worth doing. Other than that, this PR is ready for review. 

- There was a weirdness with injecting `numpy` docs into the `unumpy` docs
- I rewrote the section on saving/loading `numpy` arrays of `UFloat`s
- I change the CI configuration a little bit.

jagerber48:
Yeah I'm going to leave it at this. I think arrays of `ufloat` need a dedicated method for "nice" printing and I don't want to work on that just to get the docs looking pristine with pristine doctests. we can work on that as we improve the `unumpy` interface.

.github/workflows/python-package.yml
newville: @jagerber48 OK, that sounds good to me.

uncertainties/unumpy/core.py
newville: agreed

newville (approved):
@jagerber48 Thanks!  This all looks very nice to me.

newville:
@jagerber48 Thanks for doing this.   
Github is suggesting a "Rebase and merge".   Is that acceptable?

newville:
@jagerber48 Oh, you are right: "allow merge requests" is not checked as a valid merge method.  I'm sorry, but I do not recall that conversation

I would slightly prefer "regular merge" here too, though "squash and merge" would be OK. 
Anyway, happy to merge with whatever method is best.

wshanks:
My preference is for squash and merge because it keeps the history clean. I don't feel strongly and I am not committing as much you all so I defer to your preferences though. If a PR has unrelated changes that should be kept as separate commits, I think rebase and merge is better than a plain merge if it works. If there are merge conflicts in the middle of the branch that get resolved later, the rebase option doesn't work though.

jagerber48:
Let's just squash merge this and the other PRs to get them in. We can discuss the merge strategy elsewhere. I'll merge this one and some other approved PRs tomorrow if other folks haven't by then.

PR #303, opened 2025-04-18
Release 3.2.3

jagerber48:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,46 +4,54 @@ Change Log
 Unreleased
 ----------
 
+3.2.3   2025-April-18
+-----------------------
+
 Changes
 
-- Changed how `numpy` is handled as an optional dependency. Previously,
+- Changes how `numpy` is handled as an optional dependency. Previously,
    importing a `numpy`-dependent function, like `correlated_values`,
    without `numpy` installed would result in an `ImportError` at import
    time. Now such a function can be imported but if the user attempts to
    execute it, a `NotImplementedError` is raised indicating that the
    function can't be used because `numpy` couldn't be imported.
-- Refactored the implementation for the calculation of the derivatives of the power
-   function and improves the corresponding testing.
+- Refactors the implementation for the calculation of the derivatives of
+   the power function and improves the corresponding testing.
 
 Adds:
 
-- Added a small benchmarking suite to CI to guard against absolute performance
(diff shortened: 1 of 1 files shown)


jagerber48:
After this PR is merged I can create a github release that copies over the changelog.

I guess I'll need to push to `pypi` as well. Should I just do that manually from my local computer? We should set up a github action to do that when we make github releases.
Here's how `sciform` does it:
https://github.com/jagerber48/sciform/blob/main/.github/workflows/python-publish.yml

Getting the `pypi` credential right always seems a bit tricky...

newville (approved):
Nice - clear and comprehensive.

andrewgsavage:
> After this PR is merged I can create a github release that copies over the changelog.
> 
> I guess I'll need to push to `pypi` as well. Should I just do that manually from my local computer? We should set up a github action to do that when we make github releases. Here's how `sciform` does it: https://github.com/jagerber48/sciform/blob/main/.github/workflows/python-publish.yml
> 
> Getting the `pypi` credential right always seems a bit tricky...

yea it would be good to do that before 4.0.0 to make it easier to fix anything quickly.

newville:
@andrewgsavage @jagerber48 @wshanks @lebigot any objections to merging this, then tagging and releasing version 3.2.3?

I'd be happy for anyone to merge this.  
If no one else does that, I'll plan on merging this on Monday unless there are objections.

jagerber48:
@newville That sounds fine. It would be a "nice-to-have" to get the `pypi` release CI in to do this release so it could have a run before the `4.0.0` release but that seems like unnecessarily slowing down `3.2.3`. I think we should merge this and release `3.2.3`. 

@newville after you merge it in let us know if you publish the release to `pypi` or if you want me to do that.

Maybe `3.2.4` can have the `pypi` release CI code. My experience with the `pypi` release code is that it's not straightforward to get it right, and getting it right may result in a couple of patch releases while you mess with things. It might be nice to get that in before `4.0.0` so that we don't end up on `4.0.3` or something just because we were messing with the `pypi` release CI. But even that wouldn't be a big deal.

newville:
@jagerber48 OK, Thanks for all the hard work here.
 
Version 3.2.3 tagged and released on Github and pushed to PyPI.  

Do I recall correctly that readthedocs should automatically update based on Github releases?

PR #305, opened 2025-04-22
Increase the stacklevel of the std_dev==0 warning

wshanks:
With this change, the warning will show the caller's location instead of the location in `ufloat()` where the warning is generated.

I ran into some std_dev==0 warnings in a project I work on, partly from lmfit (like reported [here](https://github.com/lmfit/lmfit-py/issues/999)) and partly local to the project (reporting parameter sets uniformly as `UFloat` whether or not they have uncertainty). The `ufloat()` calls are made many levels deep in the code so it was hard to tell where they were coming from (I put a break point on the warning with a debugger). I think incrementing the stack level so the caller is blamed for the warning is helpful.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Change Log
 Unreleased
 ----------
 
+- Increases the stacklevel of the warning about `std_dev==0` so that it blames
+  the caller rather than blaming the `ufloat()` function.
+
 3.2.3   2025-April-18
 -----------------------
 
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -1021,7 +1021,10 @@ def ufloat(nominal_value, std_dev=None, tag=None):
        and `tag` which match the input values.
     """
     if std_dev == 0:
-        warn("Using UFloat objects with std_dev==0 may give unexpected results.")
+        warn(
+            "Using UFloat objects with std_dev==0 may give unexpected results.",
+            stacklevel=2,
+        )
     return Variable(nominal_value, std_dev, tag=tag)
 
 


newville (approved):
@wshanks Thanks -- I agree. 

It seems like it is always a bit challenging (to me, anyway!) to set Warning levels, but I think this gets it right.

jagerber48:
This looks good to me, thanks! Once this is merged in should we rapidly release `3.2.4` since the original change is already impacting other projects? I can't tell if changing this stack level will resolve issues the other packages are seeing or if it will require changes on downstream projects either way...

wshanks:
@jagerber48 In the default case (not changing Python's warnings settings), I don't think this change will make the warning more or less disruptive. It should make the warning a little easier to understand because the default warning message shows the code line where the warning was emitted from but not the full traceback like an exception. So with this change instead of seeing a message like:

    D:\Lib\site-packages\uncertainties\core.py:1024: UserWarning: Using UFloat objects with std_dev==0 may give unexpected results.
      warn("Using UFloat objects with std_dev==0 may give unexpected results.")

you should see a message like:

    /lib/python3.13/site-packages/lmfit/parameter.py:576: UserWarning: Using UFloat objects with std_dev==0 may give unexpected results.
      uvars[par.name] = ufloat(par.value, stderr)

which helps you know where to look to try to address the warning.

In projects that I work with, I like to put `warnings.filterwarnings("error", module=<package>)` in the test code so that all warnings attributed to the package's code get turned into exceptions. Before this PR, the std_dev==0 warning would not be turned into an exception because the module for the warning would be `uncertainties.core` rather than the module in my project calling `ufloat()` but with this PR it could now raise an exception if the project was passing `std_dev=0`. So this PR could be more disruptive in that sense but it would also be what I would want.

I think it is good to make another release but not too urgent. Maybe we could try getting the GitHub PyPI release in place and use the 3.2.4 release as the first real test if there were interest in setting that up soon.

newville:
@andrewgsavage I agree.  Let's try to set that up.

wshanks:
Recently I set up trusted publishing for another small project ([this one](https://github.com/wshanks/pytest-unittest-id-runner)) and it worked okay for me. I used the template [here](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi). There is also a [discussion](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) in the Python Packaging User Guide. The template I used triggers the PyPI upload off of creating a GitHub Release but we could trigger off a tag instead. The template in the Python Packaging User Guide does it that way (and even automates creating the GitHub release from the tag and PyPI upload).

PR #308, opened 2025-04-29
Pypi publish ci

### LINKED ISSUE #307: pypi publish CI
We should set up CI to automatically publish to pypi

`sciform` does it using this github workflow: https://github.com/jagerber48/sciform/blob/main/.github/workflows/python-publish.yml
This workflow trigger whenever a github release is made. My user workflow is: (1) Merge in the release branch to main (2) create a github release, copying over the changelog, and creating a new tag as part of the release. (3) the github action runs a publishes to pypi.

`pint` uses https://github.com/hgrecco/pint/blob/master/.github/workflows/ci.yml
It looks like this workflow triggers whenever a tag is pushed

I would need to brush up on details. I think I prefer the first strategy because it's a little more controlled. e.g. it's easier to accidentally push a tag than it is to accidentally make a github release. FWIW, it also allows for tags in the repo which aren't pushed to pypi.

_______
When setting this up for `sciform` I kind of remember the crux of the problem being setting up the pypi token/authentication/etc. correctly with github.

jagerber48:
- [ ] Closes #307

 This PR:
- Adds a github workflow for publishing to pypi based off of the template at https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi
- Renames the existing test CI workflow

To make the action work, in the pypi settings, I set up github as a trusted publisher in pypi. Also, per recommendation, in the github repo settings, I set up a github workflow environment that this workflow will run in.


jagerber48:
This PR is ready for review. Not sure if it's ready for approval or not. Excited to get other folks opinions of what I have here.

newville (approved):
This looks fine to me. I have not done this sort of automated  publishing of releases in other projects, so I would defer to others on whether this is the preferred method in detail.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+      name: pypi-publish
+      url: https://pypi.org/project/uncertainties/
+    steps:
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+      - name: Publish release distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@6f7e8d9c0b1a2c3d4e5f6a7b8c9d0e1f2a3b4c5d
```
wshanks, line 44: Where did this hash come from? I couldn't find it in the https://github.com/pypa/gh-action-pypi-publish repo. Should we just use `v1`?

wshanks (approved):
Looks good to me. I was curious about the sha used for the last publish step.

.github/workflows/publish.yml
jagerber48: Happy to use the release. I don't recall where the SHA came from and I didn't fully understand why the SHA is better than the release..

wshanks (approved):

.github/workflows/publish.yml
wshanks: The part after the `@` is a git reference. The nice thing about a sha is that it can't easily be changed. Even if you don't trust the upstream, it takes some effort to change the code in a malicious way and have the hash come out the same. I think we can trust pypa though. By using `v1` we are pointing to a branch rather than a tag. That means if they make changes we will get them automatically without needing to change the number in the file. We just hope that they switch to `v2` before making any breaking changes.

Issue #273, opened 2024-12-07
Clean up tests


jagerber48:
#262 makes changes to the backend for some core functionality and also has some API changes. As a result it comes with decently large changes to the test suite. That is, some tests are actually changed.

However, many of the tests are written in ways that could be heavily modernized while still retaining the same test cases. I'm am restraining from doing much of this in #262 (though I am doing some). It would be beneficial to go through all tests and "clean them up". It would improve readability and make future maintenance easier.

PR #311, opened 2025-05-01
Improve "is close" testing

jagerber48:

Description:

- `uncertainties` still uses some custom "is close" checks. These checks should be delegated to `math.isclose` and `numpy.isclose`.
- The `uarrays_close` function doesn't actually check that the `UFloat`s in the arrays are close. Rather it checks that the nominal values are close and the std_devs are close (doesn't ensure correlation)
- The `uarrays_close` also uses the `to_affine_scalar` function. With discussions like #295 and others we are moving towards more tight type control. Some of the changes going into `4.0.0` drop dependence on `to_affine_scalar` so I would like to remove it entirely. It is not necessary in the `uarrays_close` test function so removing it there is the first step towards removing it later.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,13 +1,16 @@
-from math import isclose, isnan, isinf
+from math import isclose, isnan
 
-import uncertainties.core as uncert_core
+try:
+    import numpy as np
+except ImportError:
+    np = None
 
 
-def nan_close(first, second):
+def nan_close(first, second, *, rel_tol=1e-9, abs_tol=0.0):
     if isnan(first):
         return isnan(second)
     else:
-        return isclose(first, second)
+        return isclose(first, second, rel_tol=rel_tol, abs_tol=abs_tol)
 
 
 ###############################################################
@@ -17,35 +20,7 @@ def nan_close(first, second):
 # Utilities for unit testing
 
 
-def numbers_close(x, y, tolerance=1e-6):
-    """
-    Returns True if the given floats are close enough.
-
-    The given tolerance is the relative difference allowed, or the absolute
-    difference, if one of the numbers is 0.
-
-    NaN is allowed: it is considered close to itself.
-    """
-
-    # !!! Python 3.5+ has math.isclose(): maybe it could be used here.
-
-    # Instead of using a try and ZeroDivisionError, we do a test,
-    # NaN could appear silently:
-
-    if x != 0 and y != 0:
-        if isinf(x):
-            return isinf(y)
-        elif isnan(x):
-            return isnan(y)
-        else:
-            # Symmetric form of the test:
-            return 2 * abs(x - y) / (abs(x) + abs(y)) < tolerance
-
-    else:  # Either x or y is zero
-        return abs(x or y) < tolerance
-
-
-def ufloats_close(x, y, tolerance=1e-6):
+def ufloat_nan_close(x, y, tolerance=1e-6):
     """
     Tests if two numbers with uncertainties are close, as random
     variables: this is stronger than testing whether their nominal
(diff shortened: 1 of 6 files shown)


jagerber48:
Bump, I would like to have this approved + merged next.

newville (approved):
Happy to have @andrewgsavage review this too.

it seems that the "description" states the problem (ie, the Issue), but does not fully describe the proposed solution.  I may be missing some parts of the changes, but here are my impressions and suggestions.

Adding tolerances to `nan_close` seems like a good change. 

It seems like `ufloats_close` is basically replaced by ufloat_nan_close`, which then uses `nan_close`.  That's fine.  But, one comment on that is that the tests here really appear to put "Nan is possible" front and center (like, in the name, even). Are the tests actually encountering a lot of `NaN`s and expecting that these might be inconsistent, so we really need to test for this?

I would say that correlations should be included in the tests at least "sometimes", but maybe not "always"?  It seems like most of the tests do not include correlation.

I might be hesitant to make too many numerical assertions about what the tolerances on those correlations should be.   I would probably go with "very high tolerance" (like 0.1) -- so that the tested correlation is effectively either "very high", "medium", or "low". 

But also: if the consistency of the correlation is important, it may be better to test those explicitly. The implicit correlations between UFloats is sort of confusing to rely on without explicit tests.  

So maybe leave the default tests to not include correlations, but test the correlations when that is part of the actual test.
I think that is what the tests here do.

But, those are suggestions. 
I'll approve this, but happy to have others weigh in.

jagerber48:
@newville thanks, this all good feedback. I'm happy to address this before merging if others prefer, or we can merge as is.

> it seems that the "description" states the problem (ie, the Issue), but does not fully describe the proposed solution. I may be missing some parts of the changes, but here are my impressions and suggestions.

I'll go through the code again and update the description with the resolutions in addition to the issues.

> one comment on that is that the tests here really appear to put "Nan is possible" front and center (like, in the name, even). Are the tests actually encountering a lot of NaN`s and expecting that these might be inconsistent, so we really need to test for this?

Yeah, I agree that the amount of times `nan` is now appearing is more than necessary. The tests **rarely** encounter a NaN and I just wanted a function that covers all cases. I would be fine with removing `nan_close` and using `math.isclose` whenever possible and reserving `nan_close` for those few occasions when we are interested in testing NaN. I think right now the few NaN cases might be interspersed among non-edge case tests, hence the desire for using one function that handles both cases. But a better approach here would be to split out tests that include NaN from tests that do. Then throughout most of the non-edge case tests we just use `math.isclose` and in the edge case tests we use `nan_close` (which could actually be renamed to `nan_is_close` or `is_close_nan_support` or something.

> I might be hesitant to make too many numerical assertions about what the tolerances on those correlations should be. I would probably go with "very high tolerance" (like 0.1) -- so that the tested correlation is effectively either "very high", "medium", or "low".

Yeah I only used explicit values for tolerances when the old test already had a tolerance specified or if a higher version was needed for the test to pass...

Regarding all of your other comments about correlation: 

- Right now, in most of the tests, we are checking that nominal values are close and the two `UFloat` are close to perfectly correlated. But yes, in many cases the correlation test is implicit. My plan for `4.x` can/will involve more explicit tests for correlations. This will be as a result of two changes. First the `UFloat.uncertainty` attribute (which replace the `AffinerLinearFunc._linear_part` attribute) will have more prominence. if `x` is a `UFloat` you can access it via `x.u`. We can implement a test function to tell use `x.u` is close to `y.u` to tell us that `x` and `y` are perfectly correlated. I will also have `UFloat.covariance(self, other)` and `UFloat.correlation(self, other)` methods (and functions?) that will explicitly report the correlation between two `UFloat`. These can be used during testing for more explicit correlation comparison.

____________
With all of that said, I err on the side of merging this faster so that I can focus on the `4.x` code instead. I think this parallel work of developing the `4.x` code together with test cleanup is working well, but I think in this particular case many of these tests are going to be reworked for `4.x` anyways so it might not be worth going crazy making the tests perfect yet. Just good enough that they're not inhibiting `4.x`. #273 is still on my radar, but I imagine final work on #273 happening AFTER `4.x` release.

newville:
@jagerber48 Great -- I agree with all of that.  I suggest waiting a few days to give others a chance to comment, then merging unless there are objections.

andrewgsavage:
Looks fine to merge. In a future PR it might be helpful to add examples to the docstring, especially if we intend to make these helper functions public. 

> Yeah, I agree that the amount of times nan is now appearing is more than necessary. The tests rarely encounter a NaN and I just wanted a function that covers all cases. I would be fine with removing nan_close and using math.isclose whenever possible and reserving nan_close for those few occasions when we are interested in testing NaN. 

I like that idea

newville:
@jagerber48 @andrewgsavage Thanks! merging...

PR #312, opened 2025-05-05
Fix Sphinx copyright year substitution

cjwatson:
Using a Unicode en dash to separate years in the Sphinx `copyright` option results in unreproducible builds (in the sense of https://reproducible-builds.org/), because Sphinx's special handling of `SOURCE_DATE_EPOCH` only works when the separator is an ASCII hyphen. The Sphinx developers seem to have considered changing this and then decided against it (see
https://github.com/sphinx-doc/sphinx/pull/12738), so it seems best to adjust the configuration here.

- [N/A] Closes # (insert issue number)
- [N/A] The change is fully covered by automated unit tests
- [N/A] Documented in docs/ as appropriate

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,12 @@ Change Log
 Unreleased
 ----------
 
+Fixes:
+
 - Increases the stacklevel of the warning about `std_dev==0` so that it blames
   the caller rather than blaming the `ufloat()` function.
+- Adjusts the Sphinx configuration to allow for reproducible builds using
+  ``SOURCE_DATE_EPOCH``.
 
 3.2.3   2025-April-18
 -----------------------
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -44,7 +44,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "uncertainties"
-copyright = f"2010–{date.today().year}, Eric O. LEBIGOT (EOL)"
+copyright = f"2010-{date.today().year}, Eric O. LEBIGOT (EOL)"
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the


newville:
@cjwatson I don't understand what problem this solves.  Somebody made up some "best practice" about formatting copyright dates in documentation?

If someone has a workflow based on parsing the copyright notice of the documentation, they should be brought to an awareness about the fragility of that assumption.  Git it really good at marking and time-stamping changes to a collection of files, and tagged releases and/or pushes to PyPI should be used to decide changes in releases and their dates. 

If we are touching the copyright notice for the docs, then @jagerber48 and I should be added as authors, or perhaps just change that to "Uncertainties Developers".   

I vote to close.

newville:
@cjwatson Why would a "reproducible build" process which clearly states (100% I know what this thing is):

``` 
reproducible builds are a set of software development practices that create an independently-verifiable path from source to binary code.
```

be concerned about the copyright notice for the documentation? 

I'm sort of still leaning towards "Nope".

Anyway, the whole "reproducible build" thing is inherently uninteresting for this project, as it is pure Python.  There is no path from source code to binary -- there is nothing to verify.

cjwatson:
It is not in general sustainable or realistic to do reproducibility checks with a pile of exceptions for things that might be different, not least because small differences have a way of embedding themselves into nested file formats.  We check reproducibility by confirming that the build outputs (which in this case includes the generated documentation) are bit-for-bit identical.

By the same token, it's not sensible to have exceptions for packages that ostensibly have trivial build procedures.  Sure, the process of going from the source tree for uncertainties to a binary output might be trivial.  That does not prevent somebody from injecting code into a build machine (whether that's for wheels, .debs, RPMs, or whatever) that includes some kind of stealthy modification in the build outputs.  Reproducible builds give us a way to detect that kind of attack, even if it's only after the fact, by running independent builds and comparing the outputs.  And doing that is a lot more effective when the outputs can actually be reliably compared, without having to maintain a complex set of known exclusions.  (This is just one package among many.)

wshanks:
I was also puzzled at first because why does a Sphinx configuration affect what goes into the Python package, but looking [here](https://tracker.debian.org/pkg/uncertainties) I see that debian actually packages the Uncertainties documentation as a separate package. I can also see there that @cjwatson helps maintain it. Personally, I don't mind changing which unicode character is used for the hyphen in the Sphinx copyright to the form that some part of Sphinx already assumes.

newville:
A packaging system should have a copy of the source kit, and store some hash/checksum key of that kit to prevent alteration of the source code.  It can build and install from that source kit any way it wants, including making any patches it wants.   

Why should we have to change our code to match the assumptions of some packaging system?

jagerber48:
> Sphinx has included support for SOURCE_DATE_EPOCH for many years, which is part of its logic for deciding when to adjust the copyright option to match the current year. As you can see from https://github.com/sphinx-doc/sphinx/blob/master/sphinx/config.py#L769, this logic is sensitive to the punctuation used in the copyright option.

Yeah so this is the problem. It's a bit insane that something depends on the exact punctuation and then repos are being asked to make these punctuation changes.

One part of me says this is a minor change that we can make to make someone's life easier. But another part of me is tempted to reject this so that the right parties (sphinx and reproducible-builds) figure out a better resolution to eliminate the punctuation sensitivity.

jagerber48:
It sounds like right now the punctuation sensitivity in sphinx is retained so as not to break packages that are deliberately working around this in some way. But I guess now other package, who don't really care about this, are being asked to make changes.

cjwatson:
@jagerber48 While I don't have figures for everything using Sphinx anywhere, https://tests.reproducible-builds.org/debian/issues/unstable/copyright_year_in_documentation_generated_by_sphinx_issue.html suggests that this is a pretty rare issue at least in the set of packages in Debian.  In fact, although I haven't looked very deeply, the other package listed there seems to be unreproducible for some other reason and not this one; so from what I can tell there is _one_ significant package affected by this problem, certainly not thousands or probably even very many.

I noticed this because I'm on the Debian Python packaging team, which deals with the bulk of the Python modules that have been packaged in Debian for one reason or another (in this case probably because it's an indirect dependency of a few applications that somebody wanted to make available in our distribution).  There are a lot of these and we generally rely on dashboards to keep things manageable.  One of the things on our dashboard is whether packages build reproducibly; the great majority do, but a few such as this one don't.

@wshanks Yes, we often package documentation for various reasons.  (Though I wasn't the one who set that up in this case; I've just been keeping the package up to date.)

@newville We can and do apply patches, of course; in fact I already applied this patch to our package, so this PR isn't blocking me.  Most upstream maintainers prefer it if we send patches upstream wherever possible, though; and often similar problems affect more than one downstream distributor.

I don't want this to be an adversarial business.  It's obviously up to you whether you take this patch; I can't really spend ages trying to persuade the maintainers of one package to take a one-character patch when we have thousands of packages to deal with!  But it does seem as though perhaps the `copyright` option as currently written doesn't actually quite make sense for how this package is currently being maintained, so maybe you want to adjust it in some other way even if you aren't happy with this change.

wshanks:
> this is a minor change that we can make to make someone's life easier

This was my first reaction. I did not actually know en dash existed, only em dash and hyphen. Looking up en dash, I see that one of its main uses is to separate dates, so I am less inclined to switch to hyphen, though I will personally just keep using hyphen rather than en dash in my own typing. It seems like Sphinx should reconsider its behavior and offer an explicit opt out rather than behaving differently for hyphen and en dash.

I could see an argument for hard-coding the date. It is odd to think of someone else re-running the Sphinx code years from now and extending the copyright date. It is a chore to update it manually though.

andrewgsavage:
> > this is a minor change that we can make to make someone's life easier
> 
> This was my first reaction.

That's my thoughts too. This change is practically unnoticable. 

> I could see an argument for hard-coding the date. It is odd to think of someone else re-running the Sphinx code years from now and extending the copyright date. It is a chore to update it manually though.

Absolutely. I wonder how copyright dates are handled in other larger packages.


I think we should either 
- merge this
- handle copyright dates in some other way, eg hard-coding
- one of us puts a patch in to fix this in the other modules  (this seems more effort than it's worth)

jagerber48:
I vote +1 to merge this with some minor changes. @cjwatson has taken time clearly spelling out the case for this minor change. I don't feel a strong need to "send a message" with our response to this PR.

I would appreciate a minor message in the changelog pointing out how this is in response to a quirky behavior in sphinx. We can refer to https://github.com/sphinx-doc/sphinx/pull/12738 in the changelog even.  I think the discussion in this PR should be referenced also. @cjwatson if you don't want to do that then we can update the changelog ourselves when we make the next release, but the details are fresh in everyone's heads now so it will be easier. If need be, we can clean up the formatting etc. ourselves to conform to our loose conventions at release time.

But, I do think the sphinx behavior is essentially a bug and I'm frustrated that we had to have this conversation here and spend time/energy so that other packages can somehow rely on this buggy behavior as "backwards compatible" feature. So I would appreciate if an issue was opened with sphinx re-visiting https://github.com/sphinx-doc/sphinx/pull/12738. I would be happy to put a comment on such an issue expressing this point.

cjwatson:
Maybe the best thing to do first is to loop in @jayaddison on this to see if I've missed anything in my analysis, or if they think it's likely that Sphinx might want to change anything about its handling of copyright years in light of this PR.

jayaddison:
Hello - Sphinx is doing its best to enable reproducibility for documentation projects with programmatic copyright notices; possible, as here, because `conf.py` is dynamically-evaluated.  As you're all no doubt aware, wrangling dynamically-evaluated notices to stable/reproducible outputs is never going be perfect, so it's a best-effort approach.

We've been aware of this limitation/problem, and from Sphinx v8.1.0 onwards there is a `%Y` substitution pattern that can be placed into static/constant `copyright` / `project_copyright` config strings.  **This is what I would generally recommend in this case** - it should allow you to remove any year-calculation code, while retaining dynamic output.  See the documentation at: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-copyright

(note regarding sphinx-doc/sphinx#12738: I don't have any magic tooling to assess the impact of changes in Sphinx itself on downstream projects (some of which will be closed-source) -- but the reason I pushed back on that was that I'm aware of at least one (open-source) project that intentionally did the opposite - they _began_ using em-dash precisely to avoid Sphinx's substitution logic.  instead of trying to pick a winning side, I wanted to accomodate existing projects and find a likely-mutually-acceptable path forward)

jayaddison:
Mea culpa: I've tested my own recommendation, and using the `%Y` substitution -- although I do still recommend it -- does not intrinsically use `SOURCE_DATE_EPOCH`; that remains reliant on the pattern-matching substitution, and therefore ASCII-hyphen.

I'm sorry about that; I'd forgotten how these features related to each other.

newville:
I am still -1 -- willing to be out-voted.  

The current code is not broken in a way that we see in our tests or usage.  This is a patch for their code and assumptions, not ours.   There are no guarantees that we won't "break" this at some point in the future.  

We are not obligated to support others building the docs or guarantee that building the docs will yield byte-for-byte compatible results.  

As an example: the lmfit-py documentation uses the Sphinx gallery package to generate results that includes the run time to milliseconds and auto-generated images (see https://lmfit.github.io/lmfit-py/examples/example_fit_with_inequality.html#sphx-glr-examples-example-fit-with-inequality-py for example).  It is not byte-level reproducible.   The assumption that rendered docs must be byte-level reproducible is flawed. 

Again, willing to be out-voted.

jagerber48:
@jayaddison 
> but the reason I pushed back on that was that I'm aware of at least one (open-source) project that intentionally did the opposite - they began using em-dash precisely to avoid Sphinx's substitution logic. instead of trying to pick a winning side, I wanted to accomodate existing projects and find a likely-mutually-acceptable path forward)

But this seems odd. Having code reliant on whether a certain character is an en-dash or a hyphen is very fragile and it seems a bit crazy to have or support any code that is reliant in that way. I don't fully understand the `SOURCE_DATE_EPOCH` feature and how it is utilized, but couldn't there be an explicit configuration option in `conf.py` to either opt in or out of usage of that feature? If such a feature existed I wouldn't hesitate to change our `conf.py` to opt in as an override of the en-dash opt out if that is maintained for some temporary period of time. But changing this unicode character so that other people code works feels frustrating.

@newville 
> As an example: the lmfit-py documentation uses the Sphinx gallery package to generate results that includes the run time to milliseconds and auto-generated images (see https://lmfit.github.io/lmfit-py/examples/example_fit_with_inequality.html#sphx-glr-examples-example-fit-with-inequality-py for example). It is not byte-level reproducible. The assumption that rendered docs must be byte-level reproducible is flawed.

This is a good point. I fully agree with this. Especially 

> The current code is not broken in a way that we see in our tests or usage. This is a patch for their code and assumptions, not ours. There are no guarantees that we won't "break" this at some point in the future.

Maybe I haven't yet been burned enough by other people's demands on the codebase I maintain, but for this one I'm willing to just pass it through to make them happy. Obviously if another reproducible build issue comes up I would be less willing to comply for tenuous justification...

wshanks:
The [Wikipedia entry for en dash](https://en.wikipedia.org/wiki/Dash#En_dash) lists connecting ranges as the first use case for en dash, so the current usage is valid. On the other hand, even the Wikipedia entry notes that some style guides favor hyphen over en dash:

> Preference for hyphen vs. en dash in ranges varies. For example, the [APA style](https://en.wikipedia.org/wiki/APA_style) (named after the American Psychological Association) uses an en dash in ranges, but the [AMA style](https://en.wikipedia.org/wiki/AMA_Manual_of_Style) (named after the American Medical Association) uses a hyphen

So changing it is not a complete violation of punctuation convention.

I agree though that requiring this kind of minor change does not seem like the right way to achieve something like reproducible builds because it does not scale well , but I know reproducible builds is a long-running effort and probably anything I could come up with on the fly has already been considered.

It seems like we have a -1 vote and a +1 vote and then two ambivalent votes leaning towards +1, so we will probably merge this. If it seemed like we would vote it down, I would suggest we make the change to `%Y` and perhaps @jayaddison's [suggestion](https://github.com/orgs/sphinx-doc/discussions/13526) will make that work with reproducible builds in the future. Perhaps we should still change to `%Y` as well any way (it just requires setting the minimum Sphinx version to 8.1; the readthedocs build is using 8.2 currently).

jagerber48 (approved):
Approved. Let's just clear this out. Will merge in the next couple days unless someone else does.

PR #316, opened 2025-05-21
mark __bool__ as deprecated

### LINKED ISSUE #315: Should `UFloat.__bool__` be removed?
Previously, if `x = ufloat(1, 0.1)` then `x.__bool__()` was equivalent to `x == 0`. This was in turn equivalent to `x.n == 0 and x.s == 0`. However, the resolution to #310 is looking like it is going to eliminate equality comparison between `UFloat` and `float` (or it will always return `False` or something). This leaves us with the following options for `UFloat.__bool__`:

1. Remove `UFloat.__bool__`. Thinking of `UFloat` as a random variable, it doesn't make sense to talk about a boolean value for a `UFloat`.
2. Define `__bool__` to be `x.n == 0 and x.s == 0`. this will be backwards compatible
3. Define `__bool__` to be `x.n == 0`

These are in order of my preference. If we decide to remove it we should include a deprecation warning for it right away.

jagerber48:
- [x] Closes #315

Note

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,7 +58,7 @@ Deprecates:
    `floor`, `fmod`, `frexp`, `ldexp`, `modf`, `trunc`. The following
    `AffineScalarFunc`/`UFloat` methods are marked as deprecated:
    `__floordiv__`, `__mod__`, `__abs__`, `__trunc__`, `__lt__`,
-   `__le__`, `__gt__`, `__ge__`.
+   `__le__`, `__gt__`, `__ge__`, `__bool__`.
 
 3.2.2   2024-July-08
 -----------------------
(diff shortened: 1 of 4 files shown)


wshanks:
I think it might be better to put in the warning by hand for this case. Removing `__bool__` will result in `UFloat` falling back to `object.__bool__`. So `bool(ufloat)` will still work but will handle 0 differently. Or we could not warn and just put the change in the release notes.

jagerber48:
That's a good thought. All of the other deprecations I think are going to result in `AttributeError` or something but this one will be a silent behavior change in the end. I'll put a custom warning.

jagerber48:
@wshanks ok I added a custom warning for `__bool__` I tried to get the stacklevel right but I might have missed it again! Let me know if it looks good.

wshanks (approved):
All looks good to me!

PR #318, opened 2025-05-25
Mark matrix support as deprecated

### LINKED ISSUE #224: Review numpy API coverage
Since Uncertainties has not been updated much recently, there has been time for the functions in the unumpy module to drift relative to upstream numpy. Following on from the need to look at the numpy support because of numpy 2 (#222), it would be good to review the current unumpy interface to make sure the current functions still match numpy's versions (no change in function signature) and no functions added to numpy that a user would expect are missing. Besides that there are a couple known action items:

* numpy 2 introduced aliases for inverse trigonometric functions like `acos` for `arccos`. It would be good for unumpy to include similar aliases.
* Review if `unumpy.matrix` could be updated/deprecated. The corresponding [numpy.matrix](https://numpy.org/devdocs/reference/generated/numpy.matrix.html) has been documented as not recommended and possibly to be removed for a long time now. However, the numpy developers did not take the opportunity to remove it with numpy 2.0 and likely numpy 3.0 will not happen for a long time. Still, `unumpy.matrix` could perhaps migrate to deriving from a standard array instead of `numpy.matrix`.

jagerber48:

This code addresses but won't close #224.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,13 @@ Adds:
 - Adjusts the Sphinx configuration to allow for reproducible builds using
   ``SOURCE_DATE_EPOCH``.
 
+Deprecates:
+
+- Support for interoperability between `uncertainties` `AffineScalarFunc` scalar objects
+   and `np.matrix` objects is now marked as deprecated. According to the
+   `numpy documentation <https://numpy.org/doc/stable/reference/generated/numpy.matrix.html>`_
+   usage of the `np.matrix` class is no longer recommended.
+
 3.2.3   2025-April-18
 -----------------------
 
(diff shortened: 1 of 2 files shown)


jagerber48:
If we don't need tests for the deprecation warnings then this is ready for review

jagerber48:
Yeah, I'd rather just not mess with documentation much. The documentation is going to be pretty dramatically edited for 4.0 so I just don't want to put much effort into it. 

I would say if there's a worst case scenario where we release 4.x and there's a problem for some time and users are pinning to <4.0 we can release a new 3.x version with updated documentation. or something like that...

If others have the other point of view let me know and I'm happy to take the time to update the docs. Otherwise let's merge this.

wshanks:
I agree with not putting much effort into the docs but just adding some warnings seems reasonable to me: https://github.com/lmfit/uncertainties/pull/322.

PR #325, opened 2025-06-03
Remove matrix support

jagerber48:

### DIFF
--- a/tests/test_ulinalg.py
+++ b/tests/test_ulinalg.py
@@ -33,23 +33,9 @@ def test_list_inverse():
     mat_list_inv_numpy = numpy.linalg.inv(mat_list)
     assert type(mat_list_inv) == type(mat_list_inv_numpy)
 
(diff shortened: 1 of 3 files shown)


jagerber48:
@andrewgsavage @newville @wshanks this is a PR into `main` for `4.0.0` but I would appreciate getting reviews on it since I don't think the changes to `unumpy` have been discussed elsewhere. While I very much think it is the right move to drop matrix support, I could see some users being inconvenience by losing the ability to define a matrix of `UFloat` and do `mat.I`.

jagerber48:
@wshanks thanks for the feedback. The two function you mention get rewritten in the `feature_linear_combo_refactor` branch and don't use `np.matrix`.

Also I'm leaving documentation out of these intermediate PRs. Shortly I'm going to be doing a thorough look over all the documentation to get it in line with the many changes that are happening.

PR #339, opened 2025-07-14
Release announcement

jagerber48:

Here is a draft of the release announcement. Please suggest or make changes to the announcement.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
 The `source code`_ for the uncertainties package is licensed under the `Revised
 BSD License`_.  This documentation is licensed under the `CC-SA-3 License`_.
 
+Version 4.0 Coming Soon
+=======================
+
+The :mod:`uncertainties` team is working on developing a new major release.
+This release modernizes the :mod:`uncertainties` codebase and simplifies the underlying
+architecture used for uncertainty propagation. These changes greatly improve the
+maintainability of this code base which helps guarantee its functionality and support
+into the future. At the heart of the architecture change is the idea that
+:class:`UFloat` objects should closely resemble mathematical random variables. Here are
+some specific benefits from the architecture update.
+
+- It will now be possible to hash :class:`UFloat` objects in a well-controlled way.
+- :class:`UFloat` objects will exhibit more intuitive behavior under copying and
+   pickling.
+- There is a clear path towards implementing correlation-preserving serialization and
+   deserialization of :class:`UFloat` objects.
+- Previously when users generated a number with uncertainties, the resulting object may
+   have been a :class:`Variable` or :class:`AffineScalarFunc` instance depending on how
+   the object was created. Now all numbers with uncertainty are uniformly represented by
+   :class:`UFloat` instances. :class:`AffineScalarFunc` remains as a legacy name for
+   :class:`UFloat`.
+
+In addition to reformating the underlying architecture, this update makes a number of
+changes and simplifications to the API to improve consistency. Some of these API change
+break backwards compatibility.
+Notable changes are
+
+- Elimination of the :class:`Variable` class.
+- Elimination of the :attr:`AffineScalarFunc.derivatives` property.
+- Replacing the :func:`AffineScalarFunc.error_components()` method with the
+   :attr:`UFloat.error_components` property
+- Elimination of a number of :class:`UFloat` instance methods and :mod:`umath` module
+   functions which do not make sense for random variables such as comparison methods
+   and modulo methods and functions.
+
+There will continue to be support for the latest version 3 release as the community
+transitions to the new version.
+
+
 .. _uncertainties package: https://pypi.python.org/pypi/uncertainties/
 .. _error propagation theory: https://en.wikipedia.org/wiki/Propagation_of_uncertainty
 .. _soerp: https://pypi.python.org/pypi/soerp
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -58,7 +58,7 @@ some specific benefits from the architecture update.
    :class:`UFloat`.
 
 In addition to reformating the underlying architecture, this update makes a number of
-changes and simplifications to the API to improve consistency. Some of these API change
+changes and simplifications to the API to improve consistency. Some of these API changes
 break backwards compatibility.
 Notable changes are
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -46,11 +46,11 @@ into the future. At the heart of the architecture change is the idea that
 :class:`UFloat` objects should closely resemble mathematical random variables. Here are
 some specific benefits from the architecture update.
 
-- It will now be possible to hash :class:`UFloat` objects in a well-controlled way.
-- :class:`UFloat` objects will exhibit more intuitive behavior under copying and
-   pickling.
-- There is a clear path towards implementing correlation-preserving serialization and
-   deserialization of :class:`UFloat` objects.
+- :class:`UFloat` objects will now be immutable and hashing them will follow the python
+   data model rule that two objects which are equal should have the same hash.
+- Previously, :class:`UFloat` instances did not preserve correlations upon copying,
+   pickling, or any other form of serialization. Now :class:`UFloat` instances will
+   preserve correlations upon copying and pickling.
 - Previously when users generated a number with uncertainties, the resulting object may
    have been a :class:`Variable` or :class:`AffineScalarFunc` instance depending on how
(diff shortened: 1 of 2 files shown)


jagerber48:
This PR is ready for review. I expect direct or suggested revisions/edits.

jagerber48:
@newville @wshanks @andrewgsavage this PR is ready for review. Specifically looking for feedback on the wording/topics covered in the announcement.

> - There is a clear path towards implementing correlation-preserving serialization and
   deserialization of :class:`UFloat` objects.
   
This one is a little interesting. This is honestly one of the things I'm most excited about but it isn't actually implemented in 4.0. Though It could be implemented in 4.1 or 4.2 or something. Not sure if it makes sense announcing it here since it's won't be released in 4.0, but I think it's exciting so I wanted to mention it. Open to feedback.

jagerber48:
@newville @wshanks @andrewgsavage bump again, looking for feedback on the wording. If I don't hear anything in a week I'll probably move towards merging it myself, though I would prefer not to.

doc/index.rst
```diff
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
...
+
+The :mod:`uncertainties` team is working on developing a new major release.
+This release modernizes the :mod:`uncertainties` codebase and simplifies the underlying
+architecture used for uncertainty propagation. These changes greatly improve the
+maintainability of this code base which helps guarantee its functionality and support
+into the future. At the heart of the architecture change is the idea that
+:class:`UFloat` objects should closely resemble mathematical random variables. Here are
+some specific benefits from the architecture update.
+
+- It will now be possible to hash :class:`UFloat` objects in a well-controlled way.
```
wshanks: I was wondering what "well-controlled" meant here. How about being more specific like this?

```suggestion
- Hashing :class:`UFloat` objects will now follow the rules of the Python data model (objects that compare equal have the same hash value).
```

doc/index.rst
```diff
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
...
+- :class:`UFloat` objects will exhibit more intuitive behavior under copying and
+   pickling.
+- There is a clear path towards implementing correlation-preserving serialization and
+   deserialization of :class:`UFloat` objects.
```
wshanks: Pickling is part of serialization so maybe it should be moved from the first bullet to the second. The bullet about copying could be more specific. I think it wants to say "correlations are preserved across copies"?

doc/index.rst
```diff
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
...
+break backwards compatibility.
+Notable changes are
+
+- Elimination of the :class:`Variable` class.
+- Elimination of the :attr:`AffineScalarFunc.derivatives` property.
+- Replacing the :func:`AffineScalarFunc.error_components()` method with the
+   :attr:`UFloat.error_components` property
+- Elimination of a number of :class:`UFloat` instance methods and :mod:`umath` module
+   functions which do not make sense for random variables such as comparison methods
+   and modulo methods and functions.
```
wshanks: I think it is helpful to note that these deprecations are covered in the 3.2 series change log and with deprecation warnings in the code.

doc/index.rst
```diff
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
...
+   the object was created. Now all numbers with uncertainty are uniformly represented by
+   :class:`UFloat` instances. :class:`AffineScalarFunc` remains as a legacy name for
+   :class:`UFloat`.
+
+In addition to reformating the underlying architecture, this update makes a number of
+changes and simplifications to the API to improve consistency. Some of these API changes
+break backwards compatibility.
+Notable changes are
+
+- Elimination of the :class:`Variable` class.
```
wshanks: Have we laid out and agreed on a set of breaking changes already? I like the direction of #262 in general but there may be places to put in shims to make the transition from 3 -> 4 smoother.

For example, maybe `Variable` should also remain as an alias to `UFloat` like `AffineScalarFunc`? The main use case I see of `Variable` searching GitHub is code that does `if isinstance(x, (AffineScalarFunc, Variable)):`.

Another example is `error_components`. I wonder if it is worth breaking compatibility just to change a method to a property. Maybe a different name could be used and the method could be kept?

I haven't looked closely at the discussions recently, so I apologize if we have already settled on what the breaking changes are.

doc/index.rst
```diff
@@ -35,6 +35,45 @@ values.  For other approaches, see soerp_ (using higher-order terms) and mcerp_
...
+   deserialization of :class:`UFloat` objects.
+- Previously when users generated a number with uncertainties, the resulting object may
+   have been a :class:`Variable` or :class:`AffineScalarFunc` instance depending on how
+   the object was created. Now all numbers with uncertainty are uniformly represented by
+   :class:`UFloat` instances. :class:`AffineScalarFunc` remains as a legacy name for
+   :class:`UFloat`.
+
+In addition to reformating the underlying architecture, this update makes a number of
+changes and simplifications to the API to improve consistency. Some of these API changes
+break backwards compatibility.
```
wshanks, line 62: I think both the benefits and the breaking changes listed involve advanced usage patterns (copying, pickling, using AffineScalarFunc/Variable). Perhaps somewhere the announcement should mention that using `ufloat` to create values with uncertainties and using the non-deprecated math functions on those values will continue to work without disruption.

wshanks (commented):
The information looks reasonable to me. I put in some comments thinking about it mainly from the perspective of a user trying to parse what changes I need to do to prepare for the changes.

doc/index.rst
jagerber48: Yeah this is a good revision, thank you! I will also include the rule that hashable objects should be immutable.

doc/index.rst
jagerber48: Ah, this is helpful feedback. I was a little mixed up here with what I was thinking about. I was thinking about serializing a `UFloat` into `.json`, something which is touched on in the documentation. Previously there wasn't really a way to do that. `uncertainties==4.0.0` doesn't really provide a direct way to do this either but there is a "clear path towards implementing" it now.

However, in `uncertainties=4.0.0` you CAN already using `pickle` for serialization and deserialization.

So yeah, these two bullet points can be cleaned up. I lumped copying and pickling together because I think that's sort of how they were in the tests but I'll take your suggestion.

doc/index.rst
jagerber48: I'm fine to leave `Variable` as a legacy alias for `UFloat`. This seems like a good idea if we're keeping `AffineScalarFunc`.

> Another example is error_components. I wonder if it is worth breaking compatibility just to change a method to a property. Maybe a different name could be used and the method could be kept?

I'm not sure about this. I don't have an immediate idea for a better name but I'm open to ideas. I'm kind of in favor of taking the opportunity to make the API good and not bake in historical warts, never to be changed. Notably, keys in `error_components` used to be `Variable` instances, but now they are `UAtom` instances. It seems to me there is a good chance that if someone was using `error_components` before this change will be dramatic enough that their code is already broken without the `function/property` change. 

Happy to have more discussion on this.

wshanks: > I'm fine to leave Variable as a legacy alias for UFloat. This seems like a good idea if we're keeping AffineScalarFunc.

I realized the case of `if isinstance(x, (AffineScalarFunc, Variable))` is redundant since `Variable` is a subclass of `AffineScalarFunc` any way. Still, it doesn't hurt to keep an old alias for this case.

> Happy to have more discussion on this.

I realized you have an issue for this in https://github.com/lmfit/uncertainties/issues/336. Let me comment there.

jagerber48:
I would like to include a longer exposition somewhere in the documentation about removing the `derivatives` property. In `4.0` I can easily imagine having a section in the docs with an explanation. For this release announcement I think it suffices to just say 

> Elimination of the :attr:`AffineScalarFunc.derivatives` property.

If users want more info before `4.0` they can find these github PRs/discussions/issues.

Thoughts? Do we need more info about this decision in the release announcement?

doc/index.rst
jagerber48: I put verbiage in along these lines

jagerber48:
@wshanks @newville this PR is ready for review and merging if it looks good.

wshanks (approved):
Looks good to me!

newville:
@jagerber48 @wshanks  Great!  I will merge this too.

newville:
@jagerber48 @wshanks Does this merge now mean we are ready to change the "Unreleased" here, tagging and actually releasing 3.2.4?

Issue #338, opened 2025-07-14
Many warnings in the 3.x tests


jagerber48:
There are many warnings when the 3.x tests run because of all the deprecation warnings. In 4.x I've taken care to suppress some expected warnings related to std_dev==0. https://github.com/lmfit/uncertainties/pull/335

Actually I now realize #335 could have been applied to `master` instead of `main`. The problem was many tests are being deleted in `main` since we are dropping support for `std_dev=0` staring with `4.0`. So I didn't want to do the work of suppressing warnings on many tests that are going to be deleted.

That last comment not withstanding: Should we take care in the tests on version 3.x to suppress all of the zero std dev and deprecation warnings? In #335 I targeted regex matched warning suppression at the test function level. We could perhaps suppress `FutureWarning` at a higher level to make this easier.

If we choose yes I would like to ask if one of the other maintainers would be willing to implement this warning suppression in tests.

PR #341, opened 2025-09-07
Update testing for Python 3.14

### LINKED ISSUE #340: Test failure with Python 3.14: `tests/test_umath.py::test_math_module - AssertionError: assert ('expected a positive input',) == ('expected a ...ut, got 0.0',)`
When running the test suite with Python 3.14.0rc2, I'm seeing a single test failure:

```pytb
============================================================== FAILURES ===============================================================
__________________________________________________________ test_math_module ___________________________________________________________

    def test_math_module():
        "Operations with the math module"
    
        x = ufloat(-1.5, 0.1)
    
        # The exponent must not be differentiated, when calculating the
        # following (the partial derivative with respect to the exponent
        # is not defined):
        assert (x**2).nominal_value == 2.25
    
        # Regular operations are chosen to be unchanged:
        assert isinstance(umath_core.sin(3), float)
    
        # factorial() must not be "damaged" by the umath_core module, so as
        # to help make it a drop-in replacement for math (even though
        # factorial() does not work on numbers with uncertainties
        # because it is restricted to integers, as for
        # math.factorial()):
        assert umath_core.factorial(4) == 24
    
        # fsum is special because it does not take a fixed number of
        # variables:
        assert umath_core.fsum([x, x]).nominal_value == -3
    
        # Functions that give locally constant results are tested: they
        # should give the same result as their float equivalent:
        for name in umath_core.locally_cst_funcs:
            try:
                func = getattr(umath_core, name)
            except AttributeError:
                continue  # Not in the math module, so not in umath_core either
    
            assert func(x) == func(x.nominal_value)
            # The type should be left untouched. For example, isnan()
            # should always give a boolean:
            assert isinstance(func(x), type(func(x.nominal_value)))
    
        # The same exceptions should be generated when numbers with uncertainties
        # are used:
    
        # The type of the expected exception is first determined, because
        # it varies between versions of Python (OverflowError in Python
        # 2.6+, ValueError in Python 2.5,...):
        try:
            math.log(0)
        except Exception as err_math:
            # Python 3 does not make exceptions local variables: they are
            # restricted to their except block:
            err_math_args = err_math.args
            exception_class = err_math.__class__
    
        try:
            umath_core.log(0)
        except exception_class as err_ufloat:
            assert err_math_args == err_ufloat.args
        else:
            raise Exception("%s exception expected" % exception_class.__name__)
        try:
>           umath_core.log(ufloat(0, 0))

tests/test_umath.py:248: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (0.0+/-0,), kwargs = {}, index = 0, pos_w_uncert = [0], names_w_uncert = [], args_values = [0.0], kwargs_uncert_values = {}

    @set_doc(
        """\
    Version of %s(...) that returns an affine approximation
    (AffineScalarFunc object), if its result depends on variables
    (Variable objects).  Otherwise, returns a simple constant (when
    applied to constant arguments).
    
    Warning: arguments of the function that are not AffineScalarFunc
    objects must not depend on uncertainties.Variable objects in any
    way.  Otherwise, the dependence of the result in
    uncertainties.Variable objects will be incorrect.
    
    Original documentation:
    %s"""
        % (f.__name__, f.__doc__)
    )
    def f_with_affine_output(*args, **kwargs):
        ########################################
        # The involved random variables must first be gathered, so
        # that they can be independently updated.
    
        # The arguments that contain an uncertainty (AffineScalarFunc
        # objects) are gathered, as positions or names; they will be
        # replaced by their nominal value in order to calculate
        # the necessary derivatives of f.
    
        pos_w_uncert = [
            index for (index, value) in enumerate(args) if isinstance(value, cls)
        ]
        names_w_uncert = [
            key for (key, value) in kwargs.items() if isinstance(value, cls)
        ]
    
        ########################################
        # Value of f() at the nominal value of the arguments with
        # uncertainty:
    
        # The usual behavior of f() is kept, if no number with
        # uncertainty is provided:
        if (not pos_w_uncert) and (not names_w_uncert):
            return f(*args, **kwargs)
    
        ### Nominal values of the (scalar) arguments:
    
        # !! Possible optimization: If pos_w_uncert is empty, there
        # is actually no need to create a mutable version of args and
        # one could do args_values = args.  However, the wrapped
        # function is typically called with numbers with uncertainties
        # as positional arguments (i.e., pos_w_uncert is not emtpy),
        # so this "optimization" is not implemented here.
    
        ## Positional arguments:
        args_values = list(args)  # Now mutable: modified below
        # Arguments with an uncertainty are converted to their nominal
        # value:
        for index in pos_w_uncert:
            args_values[index] = args[index].nominal_value
    
        ## Keyword arguments:
    
        # For efficiency reasons, kwargs is not copied. Instead, its
        # values with uncertainty are modified:
    
        # The original values with uncertainties are needed: they are
        # saved in the following dictionary (which only contains
        # values with uncertainty):
    
        kwargs_uncert_values = {}
    
        for name in names_w_uncert:
            value_with_uncert = kwargs[name]
            # Saving for future use:
            kwargs_uncert_values[name] = value_with_uncert
            # The original dictionary is modified (for efficiency reasons):
            kwargs[name] = value_with_uncert.nominal_value
    
>       f_nominal_value = f(*args_values, **kwargs)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
E       ValueError: expected a positive input, got 0.0

uncertainties/ops.py:507: ValueError

During handling of the above exception, another exception occurred:

    def test_math_module():
        "Operations with the math module"
    
        x = ufloat(-1.5, 0.1)
    
        # The exponent must not be differentiated, when calculating the
        # following (the partial derivative with respect to the exponent
        # is not defined):
        assert (x**2).nominal_value == 2.25
    
        # Regular operations are chosen to be unchanged:
        assert isinstance(umath_core.sin(3), float)
    
        # factorial() must not be "damaged" by the umath_core module, so as
        # to help make it a drop-in replacement for math (even though
        # factorial() does not work on numbers with uncertainties
        # because it is restricted to integers, as for
        # math.factorial()):
        assert umath_core.factorial(4) == 24
    
        # fsum is special because it does not take a fixed number of
        # variables:
        assert umath_core.fsum([x, x]).nominal_value == -3
    
        # Functions that give locally constant results are tested: they
        # should give the same result as their float equivalent:
        for name in umath_core.locally_cst_funcs:
            try:
                func = getattr(umath_core, name)
            except AttributeError:
                continue  # Not in the math module, so not in umath_core either
    
            assert func(x) == func(x.nominal_value)
            # The type should be left untouched. For example, isnan()
            # should always give a boolean:
            assert isinstance(func(x), type(func(x.nominal_value)))
    
        # The same exceptions should be generated when numbers with uncertainties
        # are used:
    
        # The type of the expected exception is first determined, because
        # it varies between versions of Python (OverflowError in Python
        # 2.6+, ValueError in Python 2.5,...):
        try:
            math.log(0)
        except Exception as err_math:
            # Python 3 does not make exceptions local variables: they are
            # restricted to their except block:
            err_math_args = err_math.args
            exception_class = err_math.__class__
    
        try:
            umath_core.log(0)
        except exception_class as err_ufloat:
            assert err_math_args == err_ufloat.args
        else:
            raise Exception("%s exception expected" % exception_class.__name__)
        try:
            umath_core.log(ufloat(0, 0))
        except exception_class as err_ufloat:
>           assert err_math_args == err_ufloat.args
E           AssertionError: assert ('expected a positive input',) == ('expected a ...ut, got 0.0',)
E             
E             At index 0 diff: 'expected a positive input' != 'expected a positive input, got 0.0'
E             Use -v to get more diff

tests/test_umath.py:250: AssertionError
```

This is Gentoo Linux amd64.

```
$ uv pip list
Package       Version                 Editable project location
------------- ----------------------- -------------------------
iniconfig     2.1.0
numpy         2.3.2
packaging     25.0
pluggy        1.6.0
pygments      2.19.2
pytest        8.4.2
uncertainties 3.2.3.post10+gf1b7e4f4e /tmp/uncertainties
```

wshanks:
- [ ] Closes #340

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
         python -m pip install .[all]
     - name: Test source code and docs
       run: |
-        python -m pytest tests/ doc/ uncertainties/ --cov=uncertainties --cov-report=xml --cov-report=term --doctest-modules --doctest-glob="*.rst"
+        python -m pytest --cov=uncertainties --cov-report=xml --cov-report=term
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4.6.0
       with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
+        allow-prereleases: true
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
     runs-on: ${{ matrix.os }}
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Adds:
   the caller rather than blaming the `ufloat()` function.
 - Adjusts the Sphinx configuration to allow for reproducible builds using
   ``SOURCE_DATE_EPOCH``.
+- Mark Python 3.13 and 3.14 as officially supported.
 
 Deprecates:
 
--- a/doc/user_guide.rst
+++ b/doc/user_guide.rst
@@ -297,7 +297,7 @@ There is not a unique definition for ``0**0``, however python takes the conventi
 If the power operation is performed on an ``(x, y)`` pair for which ``x**y`` is
 undefined then an exception will be raised:
 
->>> x = ufloat(0.0, 0.2)
+>>> x = ufloat(0, 0.2)
 >>> y = ufloat(-3.4, 0.4)
 >>> print(x**y)
 Traceback (most recent call last):
--- a/doc/user_guide.rst
+++ b/doc/user_guide.rst
@@ -297,7 +297,7 @@ There is not a unique definition for ``0**0``, however python takes the conventi
 If the power operation is performed on an ``(x, y)`` pair for which ``x**y`` is
 undefined then an exception will be raised:
 
->>> x = ufloat(0, 0.2)
+>>> x = ufloat(0.0, 0.2)
 >>> y = ufloat(-3.4, 0.4)
 >>> print(x**y)
 Traceback (most recent call last):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,5 +67,6 @@ doc = ["sphinx", "sphinx-copybutton", "python-docs-theme"]
 all = ["uncertainties[doc,test,arrays]"]
 
 [tool.pytest.ini_options]
-testpaths = ["tests"]
+testpaths = ["tests", "doc", "uncertainties"]
+addopts = "--doctest-modules --doctest-glob=*.rst"
 doctest_optionflags = "IGNORE_EXCEPTION_DETAIL"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,3 +68,4 @@ all = ["uncertainties[doc,test,arrays]"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+doctest_optionflags = "IGNORE_EXCEPTION_DETAIL"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,8 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Programming Language :: Python :: Implementation :: Jython",
     "Programming Language :: Python :: Implementation :: PyPy",
     "Topic :: Education",
--- a/tests/test_umath.py
+++ b/tests/test_umath.py
@@ -261,6 +261,12 @@ def test_math_module():
         assert err_math_args == err_ufloat.args
     else:
         raise Exception("%s exception expected" % exception_class.__name__)
+    try:
+        umath_core.log(ufloat(0, 1))
+    except exception_class as err_ufloat:
+        assert err_math_args == err_ufloat.args
+    else:
+        raise Exception("%s exception expected" % exception_class.__name__)
 
 
 def test_hypot():
--- a/tests/test_umath.py
+++ b/tests/test_umath.py
@@ -244,14 +244,19 @@ def test_math_module():
         assert err_math_args == err_ufloat.args
     else:
         raise Exception("%s exception expected" % exception_class.__name__)
+
+    # Exception message is different for int and float so create the math
+    # exception for float rather than reusing the int one above
+    try:
+        math.log(0.0)
+    except Exception as err_math:
+        # Python 3 does not make exceptions local variables: they are
+        # restricted to their except block:
+        err_math_args = err_math.args
+        exception_class = err_math.__class__
+
     try:
         umath_core.log(ufloat(0, 0))
-    except exception_class as err_ufloat:
-        assert err_math_args == err_ufloat.args
-    else:
-        raise Exception("%s exception expected" % exception_class.__name__)
-    try:
-        umath_core.log(ufloat(0, 1))
     except exception_class as err_ufloat:
         assert err_math_args == err_ufloat.args
     else:


newville:
@wshanks Thanks.  

I think this should probably change `.github/workflows/test.yml` to drop testing 3.8 (I would vote to drop testing with 3.9 too), and add testing for 3.13.  Given the title, it probably ought to add testing for with some release candidate of 3.14.  ;).   

I do not know whether 3.14rc2 is supported with GH Actions.

wshanks:
Good suggestion, @newville 

Regarding dropping 3.8 and 3.9, I agree that it is a good time. 3.9 will go end of life soon and it's unlikely we would break something in 3.9 before then any way.

I was uncertain about adding 3.14 because my first pass at `pip install .[all]` locally failed because one subdependency didn't have a 3.14 wheel yet and the source build failed until I installed a build dependency but it turns out that all the GitHub runners have the build dependencies in their default configuration, so no extra set up was needed besides setting `allow-prereleases: true`.

The CI uncovered a second test failure that I hadn't seen locally because I was running `pytest` and not the full command line in the workflow that adds the doctests. The second test failure was similar to the first -- a change in wording of an exception message related to using 0 in a math function. To deal with that I added the `IGNORE_EXCEPTION_DETAIL` doctest option. Additionally, I added some options to pytest's configuration so that running just `pytest` by itself will run the doctests along with the rest of the tests.

newville (dismissed):
@wshanks Thanks -- this looks good to me.   Thanks for adding the test with 3.14 and catching the change in wording of the ZeroDivisionError.  That seems like a fragility of using `doctests` to me, but I'm glad there is a solution.   

And thanks for dropping 3.9 and 3.8.

newville:
@wshanks I suggest we wait a day to give others a chance to comment before merging, but this looks good to me.  Thanks.

jagerber48:
I have a few comments and questions. My disclaimer is I haven't dug too deeply into everything yet so I might be missing something that has already been stated. My apologies if so.

- The PR is more about changing CI than fixing the error in the linked issue. I think the PR title should be updated to reflect that.
- I don't know if we "officially" decided the `uncertaintes` is going to support (aka test) exactly the python supported versions. This PR basically says that that is our intention. I'm almost fine with that. The one reason I hesitate is because I've found that between ~3.8 and 3.11 there are a number of convenient typing features that are nice to use but break one of these versions. In other words, if we stop testing 3.8 and 3.9 like in this PR I think there is a good chance we introduce one of these feature and break those old versions of python. I might suggest we continue to test 3.8 and 3.9 through this PR and discuss dropping support more formally elsewhere.
- I generally like how you redid the pytest config but I'm curious what `doctest_optionflags = "IGNORE_EXCEPTION_DETAIL"
` is impacting specifically that made you want to add it? I wonder if this is related to #338 and it would be more directly helped by clearing out all the warnings the tests are raising? Maybe #338 is unrelated though.
- I do not like that changing the test from `ufloat(0, 0.2)` to `ufloat(0.0, 0.2)` fixed it. `ufloat` shouldn't have different behavior depending on if the user passes in an `int` or a `float`. The behavior should be as if we cast user input to float. In my opinion we *should* cast user input to float and see if that fixes this bug rather than changing the test to make the bug go away.
- Also I'm confused how this PR addresses issue #340. I see a new test added but I don't see any diff on the test that was failing in that issue. What is going on here? Oh, looking closer I see you are initializing `exception_class` differently. Ok I do not like it again because I'm still confused. It looks like you're priming it for a `log` function with float exception, but then the `ufloat`'s are being initialized with `int`, so why does it work now?

Before getting into this I was hoping I'd be ok with this PR just going through but I have enough issues with it I'm interested to slow things down. Basically this PR looks like a band-aid fix to address #340 but (1) because of the reasons addressed in #340, I would say this is a low priority issues and (2) we are actively in the process of "cleaning up" the code and test in this package, so I would rather give a proper fix rather than a band aid that proliferates the issues we're trying to resolve.

My main recommendation would be to refactor `test_math_module()` to use `with pytest.raises...`. That will clean things up already a little bit. I also don't think there's a need to raise the exact same exception as the math module. Rather matching a regex with `pytest.raises` will suffices. Would that be sufficient to resolve these issues?

I might recommend cleaning up the test in one PR, making sure it still passes on 3.8 - 3.12 (the versions we've been supporting) and then make the version support changes and CI changes in another PR.

Sorry for putting the brakes on a little bit for this.

wshanks:
> The PR is more about changing CI than fixing the error in the linked issue. I think the PR title should be updated to reflect that.

The scope has grown since I first opened the PR. I updated the title to reflect that now.

> I might suggest we continue to test 3.8 and 3.9 through this PR and discuss dropping support more formally elsewhere.

That is fair. I don't feel too strongly, but the tests run quickly any way so it is not much overhead to keep testing 3.8 and 3.9. We should discuss dropping older Python versions in a separate issue. Given the tests pass, I don't see a reason not to add the new Python versions here.

> I generally like how you redid the pytest config but I'm curious what doctest_optionflags = "IGNORE_EXCEPTION_DETAIL" is impacting specifically that made you want to add it?

It is unrelated to #338. Check out the Python 3.14 logs from this commit https://github.com/lmfit/uncertainties/pull/341/commits/b2c0f2f8aa28c24de2a0f1346882a21c7d146de5 right before I set the `IGNORE_EXCPETION_DETAIL` option. In short, the Python builtin exception message changed in 3.14 so the doctest can't document the output for both Python <=3.13 and 3.14 at the same time if it compares the detailed message beyond the exception type.

> I do not like that changing the test from ufloat(0, 0.2) to ufloat(0.0, 0.2) fixed it.

I am not sure what you are referencing. I didn't make this change. I added a `math.log(0.0)` in order to capture the exception message for this call to the builtin `math` module. In my early commits, I accidentally deleted one part of the test that might have looked like that change, but I restored it.

> It looks like you're priming it for a log function with float exception, but then the ufloat's are being initialized with int, so why does it work now?

I didn't trace it exactly, but I believe `ufloat` casts its argument to `float` so it goes down the float path rather than the int path. All that has changed in Python 3.14 is the format of the exception message for `math.log` which now differs for `int` vs `float` arguments where before the message was the same for both cases. It seems fair to compare the `ufloat` case to passing a float to a `math` module function. Technically though it is a deviation in behavior regarding integer arguments.

> Before getting into this I was hoping I'd be ok with this PR just going through but I have enough issues with it I'm interested to slow things down. Basically this PR looks like a band-aid fix to address https://github.com/lmfit/uncertainties/issues/340 but (1) because of the reasons addressed in https://github.com/lmfit/uncertainties/issues/340, I would say this is a low priority issues and (2) we are actively in the process of "cleaning up" the code and test in this package, so I would rather give a proper fix rather than a band aid that proliferates the issues we're trying to resolve.

I see this PR as continuing the currently supported behavior which based on the tests is to match the exception messages. I do not see matching exception messages in the documentation though so we could deem that as not within the supported behavior that is enforced by tests if we wanted to. I agree that we should not belabor this kind of detail since there is an active refactor in progress, but I think the current change is the more minimal one for the moment.

jagerber48:
@wshanks Skipping responding to stuff that sounds good to me.

> > I do not like that changing the test from ufloat(0, 0.2) to ufloat(0.0, 0.2) fixed it.
> 
> I am not sure what you are referencing. I didn't make this change. I added a `math.log(0.0)` in order to capture the exception message for this call to the builtin `math` module. In my early commits, I accidentally deleted one part of the test that might have looked like that change, but I restored it.

Looking at this line in `user_test.rst`.

https://github.com/lmfit/uncertainties/pull/341/files#diff-ea7984cb50c5c7559f66200d287189c9baee3cd4445449e1056ea701b35e9cf0L300

> 
> > It looks like you're priming it for a log function with float exception, but then the ufloat's are being initialized with int, so why does it work now?
> 
> I didn't trace it exactly, but I believe `ufloat` casts its argument to `float` so it goes down the float path rather than the int path. All that has changed in Python 3.14 is the format of the exception message for `math.log` which now differs for `int` vs `float` arguments where before the message was the same for both cases. It seems fair to compare the `ufloat` case to passing a float to a `math` module function. Technically though it is a deviation in behavior regarding integer arguments.

I guess my confusion here ties in with my previous point about the test change in `user_test.rst` so maybe you can take a look at the line I linked and re-address my confusion?

> > Before getting into this I was hoping I'd be ok with this PR just going through but I have enough issues with it I'm interested to slow things down. Basically this PR looks like a band-aid fix to address #340 but (1) because of the reasons addressed in #340, I would say this is a low priority issues and (2) we are actively in the process of "cleaning up" the code and test in this package, so I would rather give a proper fix rather than a band aid that proliferates the issues we're trying to resolve.
> 
> I see this PR as continuing the currently supported behavior which based on the tests is to match the exception messages. I do not see matching exception messages in the documentation though so we could deem that as not within the supported behavior that is enforced by tests if we wanted to. I agree that we should not belabor this kind of detail since there is an active refactor in progress, but I think the current change is the more minimal one for the moment.

Ok, I'm fine with the quick fix if you can help me understand the change in `user_test.rst`.

jagerber48:
https://github.com/python/cpython/issues/101410 Here's the cypthon issue about changing the error message on `math.log`.

See how the `cpython` tests look for testing the new exception. I think we could try something along these lines instead in the future: https://github.com/python/cpython/pull/124299/files#diff-a39b1e44e20048f5b8207fee83df00a2541a7f02716113e6243c33165e628cd5R2506

newville:
> I don't know if we "officially" decided the uncertaintes is going to support (aka test) exactly the python supported versions. This PR basically says that that is our intention. I'm almost fine with that. The one reason I hesitate is because I've found that between ~3.8 and 3.11 there are a number of convenient typing features that are nice to use but break one of these versions. In other words, if we stop testing 3.8 and 3.9 like in this PR I think there is a good chance we introduce one of these feature and break those old versions of python. I might suggest we continue to test 3.8 and 3.9 through this PR and discuss dropping support more formally elsewhere.

We have to be able to drop support for older Python versions.  Saying that we support "currently supported Python versions" is perfectly reasonable and a very common approach.  It means that new versions of `uncertainties` will no longer be tested on Python 3.8 or 3.9, and may fail.  But it doesn't break existing installations. 

It might be slightly overstated, but the alternative you proposed might be extrapolated to "discuss which versions are supported for every PR".   That's not exactly "madness", but it adds work.

I wish Python had a slower release cadence, but I also think that dropping testing of Python 3.8 should not be controversial. 

> I generally like how you redid the pytest config but I'm curious what doctest_optionflags = "IGNORE_EXCEPTION_DETAIL"  is impacting specifically that made you want to add it? I wonder if this is related to https://github.com/lmfit/uncertainties/issues/338 and it would be more directly helped by clearing out all the warnings the tests are raising? Maybe https://github.com/lmfit/uncertainties/issues/338 is unrelated though.

My initial thought was to drop using `doctest` ;).  I think that @wshanks found a decent workaround.

wshanks:
> https://github.com/python/cpython/issues/101410 Here's the cypthon issue about changing the error message on math.log.

"math.log() is a special snowflake" :slightly_smiling_face: 

> I guess my confusion here ties in with my previous point about the test change in user_test.rst so maybe you can take a look at the line I linked and re-address my confusion?

Oh, sorry, that was something I tried just out of curiosity when first seeing the doctest error and did not mean to commit. In your previous comment, I thought you were referring to a change in a `test_` module and did not think of this change because I hadn't realized I had committed it. It is not related to addressing the doctest failure. The failure is addressed just by the doctest option `IGNORE_EXCEPTION_DETAIL` that causes doctest to ignore exception messages and only compare exception types (basically it just ignores the text after the colon on exception lines). That change is kind of going the opposite way to the test module change by not checking the exact exception message but I don't think doctest is (edit) not flexible enough to support different output per Python version in a clean way.

> See how the cpython tests look for testing the new exception. I think we could try something along these lines instead in the future:

I agree that a raises context manager would be a better pattern. I think we are essentially doing the same kind of exception message test though, just with the exceptions pulled from the `math` module's equivalent exception instead of being hard-coded like the CPython case. I wonder why they use the regex version with no regex wildcards rather than a plain string match.

jagerber48:
@wshanks ok great, that `user_test.rst` change was freaking me out making me think we had different behavior depending on if the input was `int` or `float`.

This PR looks good to me now after this discussion. Thanks for taking the time. Now the PR looks like (1) add support for 3.13 and 3.14 (2) minor clean up changes to CI and (3) minor change in the tests to support this special change to `math.log` exception messages.

My guess is we could probably do something better than `IGNORE_EXCEPTION_DETAIL` in the doctests but we can worry about that elsewhere. The docs are getting pretty heavily rewritten for 4.0 anyways.

newville:
>> https://github.com/python/cpython/issues/101410 Here's the cypthon issue about changing the error message on math.log.
> "math.log() is a special snowflake" 🙂

Thank you. LOL.

jagerber48:
I think adding support for 3.13 and 3.14 deserves an entry in the changelog

wshanks:
> I think adding support for 3.13 and 3.14 deserves an entry in the changelog

Sounds good. The only changes in the PR so far had been around updating the tests to handle edge cases that came up with 3.14. Nothing was added to the code itself to support 3.13 or 3.14. Now I added an entry to the changelog along with adding the 3.13 and 3.14 trove classifiers to pyproject.toml.

By the way, in my most recent push, there was one failure [here](https://github.com/lmfit/uncertainties/actions/runs/17617077617/job/50052893572?pr=341) on the first run. I triggered a rerun so it would pass. I think I remember a similar failure recently another time but I couldn't find it in the CI log.

jagerber48 (approved):

jagerber48:
> By the way, in my most recent push, there was one failure [here](https://github.com/lmfit/uncertainties/actions/runs/17617077617/job/50052893572?pr=341) on the first run. I triggered a rerun so it would pass. I think I remember a similar failure recently another time but I couldn't find it in the CI log.

I'll open an issue where we can discuss in more detail.

edit: See #343

newville:
Happy to merge. Thanks @wshanks !

PR #348, opened 2026-01-09
Release 3.2.4

jagerber48:
- [x] Closes # (insert issue number)

We are ready to release 3.2.4. This PR updates the label in the changelog. I used 2026/1/9 since it is late tonight on 2026/1/8.

I think our release process is to merge the PR, then create a github "release" which simultaneously tags the release with the version number `3.2.4`. Then our CI will build the version and upload to pypi. I'm happy to do this tomorrow, or someone else can. @newville I think you've done it for recent  releases.


CHANGES.rst
```diff
@@ -1,8 +1,8 @@
 Change Log
 ===================
 
-Unreleased
-----------
+3.2.4   2026-January-19
```
wshanks:
```suggestion
3.2.4   2026-January-9
```

jagerber48: Lol. nice typo in a 1-line PR

wshanks (approved):
Looks good!

The --- is not quite aligned now but that is okay.

newville:
@jagerber48 @wshanks OK to merge?  I am not sure whether we have tested the `publish.yml` "Upload Python Package" Github action, but this seems like a good time to try!

wshanks:
Merging and trying to use the release workflow sounds good to me!

jagerber48:
@newville @wshanks The workflow is triggered when someone creates a "github release" of the master branch. When making the release you should also create a tag so that setuptools_scm picks out the right version number for the build that gets pushed to pypi.

jagerber48:
@newville ugh I see the CI failed. I don't know what's going on off hand. Will need to investigate... We should make an issue. I think we should push to pypi manually and get this release out so we can get on with the 4.0 release work. We can deal with the CI issue separately.

newville:
@jagerber48 @wshanks Yes, I merged this PR, then added and pushed the tag '3.2.4', then used that to draft a new release with the notes from `CHANGES.rst`.  That then triggered the workflow, which is great. 

But the workflow failed  (https://github.com/lmfit/uncertainties/actions/runs/20863138687). 

I can do `python -m build` locally.   This gives a few warnings that I believe are from `pyproject.toml`.
But `twine check dist/uncertainties-3.2.4*` passes locally and could probably be uploaded.

But, because there are a few hours left, I could delete the release for 3.2.4 and the tag `3.2.4` and try to fix `pyproject.toml`, then push that and try releasing again and see if the action then succeeds? 

Is that OK for me to try?  If the action still fails, is it OK with you all if I just do `twine upload` locally?

wshanks:
I changed the deployment protection settings which I think should avoid the error that stopped the release the first time. There was a rule set to only allow publishing the main or master branch. I believe we want to publish only from tags, so I changed the rule to that (the tag is on the master branch but it doesn't seem like that counts). So I think we could try re-running the last Action again but I am okay with deleting the tag and fixing pyproject.toml if you think that's better.

jagerber48:
I'm happy with any of the suggested options.

PR #349, opened 2026-01-09
fix license values in pyproject.toml

newville:
fixes license settings in `pyproject.toml` to comply with the latest PyPA recommendations/requirements.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Adds:
 - Adjusts the Sphinx configuration to allow for reproducible builds using
   ``SOURCE_DATE_EPOCH``.
 - Mark Python 3.13 and 3.14 as officially supported.
+- Drop support for Python 3.8
 
 Deprecates:
 
@@ -25,6 +26,7 @@ Deprecates:
    and `np.matrix` objects is now marked as deprecated. According to the
    `numpy documentation <https://numpy.org/doc/stable/reference/generated/numpy.matrix.html>`_
    usage of the `np.matrix` class is no longer recommended.
+
 
 3.2.3   2025-April-18
 -----------------------
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
     runs-on: ${{ matrix.os }}
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,14 +21,14 @@ keywords = [
     "standard deviation", "derivatives", "partial derivatives",
     "differentiation"
 ]
-license = {text = "Revised BSD License"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE.txt"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Intended Audience :: Education",
     "Intended Audience :: Other Audience",
     "Intended Audience :: Science/Research",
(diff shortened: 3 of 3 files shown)


newville:
@wshanks,  @jagerber48  This fixes the `pyproject.toml` file so that PyPI and `python -m build` does not complain, but that requires dropping Python 3.8.  FWIW, we could drop Python 3.9, but this does not do that yet.

wshanks:
@newville What do you think? Should we try restarting the 3.2.4 publish workflow one time first to see if it goes through and then merge this in preparation for 4.0? Or try to get this into a revised 3.2.4? I don't object to anything here but I just wonder about if it should make it into 3.2.4 if it is not needed for the publish job to run.

newville:
@wshanks @jagerber48 Yes, let's try it one more time.   If it fails, I can build locally and should be able to push to PyPI.

CHANGES.rst
```diff
@@ -14,6 +14,7 @@ Adds:
 - Adjusts the Sphinx configuration to allow for reproducible builds using
   ``SOURCE_DATE_EPOCH``.
 - Mark Python 3.13 and 3.14 as officially supported.
+- Drop support for Python 3.8
```
wshanks, line 17: This has to move out of the 3.2.4 section now.

wshanks (commented):
Okay, I tried switching to a rule in the `pypi-publish` environment in the Environments section that restricted publishing to tags matching the pattern `[0-9][0-9]*.[0-9][0-9]*.[0-9][0-9]*` which some things I found on the internet said was the right pattern for this (fnmatch syntax) but I think they might have been wrong. I still got the same error about 3.2.4 not being allowed to publish. Rather than fiddle with that for now I just deleted the rule to allow publishing from any branch or tag and reran the workflow again and that time it seems to have worked. If anyone is interested, they can try to put the right syntax in to limit publishing only to the tags we expect to publish.

CHANGES.rst
newville: @wshanks  where should it go, then?

jagerber48: `3.2.4` is released. We need a new unreleased section and it will eventually go into, probably, `3.2.5`. Or if this is the only change we want in `3.2.5` you can make a `3.2.5` section put it in there and we can release again today or tomorrow.

newville: Totally baffled.  Did someone push 3.2.4 to PyPI?   I did not.
The changes to `pyproject.toml` should be made at some point. 

Closing.

wshanks: Sorry for the confusion. When I asked this:

> Should we try restarting the 3.2.4 publish workflow one time first to see if it goes through and then merge this in preparation for 4.0?

I meant "should we try triggering the 3.2.4 publish workflow resulting in 3.2.4 being successfully published?" So that was what I tried and it did result in the previously identified commit (current `master`) being published as 3.2.4.

newville: OK, thanks.

jagerber48: > The changes to pyproject.toml should be made at some point.

Can we just re-open this PR to keep the issue tracked so we don't forget about it? Totally independent of 4.0 timeline, just don't want to lose track of it since you've already done work for it.

newville: Unfortunately, this didn't get put into 3.2.4.  But I will leave that up to you and @wshanks.... still a bit miffed that I started working on this untagged 3.2.4, and then3.2.4 got released despite working to fix this and waiting for comments.

wshanks: Hmm, I thought my message [here](https://github.com/lmfit/uncertainties/pull/349#issuecomment-3730419725) was pretty clear that I was asking about trying to release 3.2.4 without the changes in this PR because they were not necessary for the release of 3.2.4 that we had already agreed on releasing (GitHub releases are tied to git tags).

Over the last year, my use of Uncertainties has been very minimal, so my efforts here are just recreational because I find it fun to maintain code that other people find useful. The amount of activity on this repo that evokes a negative emotion in me too high for a recreational project. There are enough other maintainers and you are more than capable of keeping this project in good shape. I am going to withdraw from maintainership at this time.

Issue #350, opened 2026-01-15
FutureWarning: AffineScalarFunc.derivatives() is deprecated.


pmaxted:
```
>>> from uncertainties import ufloat
>>> a = ufloat(1,1)
>>> b = ufloat(1,1)
>>> a+b
/indus/pflm/miniforge3/envs/teb2/lib/python3.14/site-packages/uncertainties/core.py:526: FutureWarning: AffineScalarFunc.error_components() is currently an instance method. This method is deprecated. In a future release it will be replaced with an instance property by the same name. It will be accessed by AffineScalarFunc.error_components (with no parentheses).
  return float(sqrt(sum(delta**2 for delta in self.error_components().values())))
/indus/pflm/miniforge3/envs/teb2/lib/python3.14/site-packages/uncertainties/core.py:491: FutureWarning: AffineScalarFunc.derivatives() is deprecated. It will be removed in a future release.
  for variable, derivative in self.derivatives.items():
2.0+/-1.4142135623730951
```

uncertainties version 3.2.4
python version 3.14.2 
Running on Scientific Linux but also seen on macos for the same version of uncertainties and python version 3.11.9

newville:
@pmaxted Hm, yeah, that is unfortunate.  That deprecation warning is actually correct.  But it is slightly obnoxious to display that warning to users when the deprecated usage is used internally, and the "users' usage" (adding objects returned from `ufloat()`) is not being deprecated.

I think the best workaround is (unfortunately, again) to put

```python
import warnings
warnings.filterwarnings('ignore', module='uncertainties')
```

which sort of makes the warnings something the user must deal with, and also prevents them from seeing warnings. 

If each warning were issued only once per session, that would be much better.  This is what `warnings.filterwarnings('default', module='uncertainties')` is advertised to do, and the more explicit `warnings.filterwarnings('once', module='uncertainties')` should do that too.   I don't know why they don't, but that appears to be a common problem, so 'ignore' is needed.  My view (and from seeing similar cases with other libraries) is that this means warnings in Python have little value.

jagerber48:
Thanks for sharing this. This is unfortunate. I will look at it when I get a chance. I would hope we can suppress the warning locally so the user doesn’t have to change anything. I think we should consider yanking this release. It’s a pretty big bug if uncertainties starts spewing unhelpful warnings at the user through no fault of the user. (If the user were calling the function themselves it would be a different story.) 

[...]

PR #353, opened 2026-01-21
3.2.5a

### LINKED ISSUE #352: DEPR: self-triggered deprecation warnings in 3.2.4
ref: https://github.com/lmfit/uncertainties/issues/336

If I understand correctly, `UFloat.error_components()` is equivalent to `UFloat.derivatives`. 3.2.4 deprecated both, and unfortunately the migration path isn't straight forward:
- there is no way to write code that will work with both uncertainties 3.x and 4.x. As a result, the only way to protect users of a library that already depends on uncertainties against breaking upgrades is to introduce an upper bound `<4` on the requirement, whith is an [infamous antipattern](https://iscinumpy.dev/post/bound-version-constraints/)
- even if I set an upper bound and restrict my own usage to `UFloat.error_components()` for now, I'll still get `FutureWarnings` from it because it is wired to use `UFloat.derivatives` internally. I can ignore the warnings in my own CI, but cannot protect my users against it without resorting to an even worse upper bound (`<3.2.4`).

One way to fix both problems, would be to instead deprecate the `error_components` method *in favor* of the existing `derivatives` property.
I would also advise yanking 3.2.4 for good measure.

newville:
- [x] Closes #352. #350

This should be merged and tagged as 3.2.5

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ authors = [
 ]
 description = "calculations with values with uncertainties, error propagation"
 readme = "README.rst"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 keywords = [
     "error propagation", "uncertainties", "uncertainty calculations",
     "standard deviation", "derivatives", "partial derivatives",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,6 @@ build-backend = "setuptools.build_meta"
 write_to = "uncertainties/version.py"
 version_scheme = "post-release"
 
-
 [project]
 name = "uncertainties"
 dynamic = ["version"]
@@ -21,18 +20,17 @@ keywords = [
     "standard deviation", "derivatives", "partial derivatives",
     "differentiation"
 ]
-license = {text = "Revised BSD License"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE.txt"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Intended Audience :: Education",
     "Intended Audience :: Other Audience",
     "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: BSD License",
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
     runs-on: ${{ matrix.os }}
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 Change Log
 ===================
 
+
+3.2.5 2026-January-21
+-----------------------
+
+- change warnings about deprecation of error_components() and
+  derivates to PendingDeprecationWarning
+- Drop support for Python 3.8
+
+
 3.2.4   2026-January-9
 -----------------------
 
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -448,7 +448,7 @@ class AffineScalarFunc(object):
         warn(
             f"{self.__class__.__name__}.derivatives() is deprecated. It will "
             f"be removed in a future release.",
-            FutureWarning,
+            PendingDeprecationWarning,
             stacklevel=2,
         )
 
@@ -481,7 +481,7 @@ class AffineScalarFunc(object):
             f"be replaced with an instance property by the same name. It will be "
             f"accessed by {self.__class__.__name__}.error_components (with no "
             f"parentheses).",
-            FutureWarning,
+            PendingDeprecationWarning,
             stacklevel=2,
         )
 


jagerber48:
See https://stackoverflow.com/questions/55377831/difference-between-deprecationwarning-pendingdeprecationwarning-and-futurewarni

`FutureWarning` is the right warning here. While temping, `DeprecationWarning` and `PendingDeprecationWarning` are meant for developers of python itself, not users  of python applications.

Or maybe I'm misunderstanding and we should discuss. I would like to take a bit of time to understand the right thing to do and make sure we're doing it the right away.

pyproject.toml
```diff
@@ -21,18 +20,17 @@ keywords = [
...
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Intended Audience :: Education",
     "Intended Audience :: Other Audience",
     "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: BSD License",
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
```
neutrinoceros, line 35: I'm not seeing a `python-requires = '>=3.9'` anywhere. This would arguably be more important than removing a trove classifiers, as installers like `pip` understand the former, and don't use the latter.

pyproject.toml
neutrinoceros:
```suggestion
requires-python = ">=3.9"
```

newville:
@jagerber48  Please approve.  FutureWarning is the current warning, and is widely viewed as obnoxious because it gives a warning at every usage.  You can argue your points about which deprecation warning is best while working on 3.2.6, at any pace you want.   This PR will turn off those warnings, while preserving the rest of the changes in 3.2.4.

I did not push 3.2.4, but I am also not in favor of yanking it.  

Again, please approve.

pyproject.toml
newville: We stopped testing with Python 3.8, and we make no claims about working with Python 3.8.

pyproject.toml
neutrinoceros: @newville this is critical if you're dropping Python 3.8

newville: I could not disagree more that it is "critical".   But also, fine, we can do that.   

@jagerber48 or @andrewgsavage, please merge.

I have to admit, this is straining my enthusiasm.

jagerber48:
At my current level of understanding, which may be wrong, you're using `PendingDeprecationWarning` as a hacky workaround for a bug introduced in 3.2.4. Why not just get rid of 3.2.4 and figure out what to do? Why are you strongly against yanking 3.2.4?

newville:
@jagerber48 It adds testing and support for Python 3.13 and 3.14.  It fixes the project configuration files and fixes the uploads of releases to PyPI.

If we yank 3.2.4, we go back to 3.2.3 from last April, with no process for releasing a new version.  And no clear statement or commitment to support for Python 3.13.  All changes made since April become part of a set of mistaken changes, and someone will have to separate the usable parts from the stuff to throw away.  And you now have to re-litigate every one of those changes.     I am not in favor of that.   

This PR preserves all the other changes, and suppresses the annoying warnings, which sort of follows normal Python usage. 

Yes, Python warnings do not work well.  The available choices are 1) use `FutureWarning` and issue a warning at nearly every calculation with ufloats or 2) use the very much flawed but documented and apparently "preferred" `PendingDeprecationWarning`, which is described in the official docs as "Base category for warnings about features that will be deprecated in the future".   

Again, yes, it is very weird that this is ignored by default and is a constant source of pain, but that's what we have.

Go ahead and "try to understand this" AFTER this get approved and pushed as 3.2.5.  

That you are not in favor of that worries me greatly.

neutrinoceros:
Btw I don't understand how this is supposed to resolve #352

jagerber48:
Here's my proposal:

1. Yank 3.2.4 bringing the package back to 3.2.3.
2. keep the master branch where it's at
3. Update the code to REMOVE all the deprecation warnings on `error_components` and `derivatives`
4. Mark that 3.2.4 is yanked in the changelog
5. Write a 3.2.5 changelog which explains that all changes in 3.2.4 are preserved in 3.2.5 except the `error_components` and `derivatives` deprecations are reverted.
6. release 3.2.5, taking us from 3.2.3 up to 3.2.5 which looks like 3.2.4 but without the deprecations.

If you like we can drop 3.8 support in 3.2.5 but that change is ENTIRELY orthogonal to this deprecations issue and I think should be ignored for now.
____
This proposal addresses
> And no clear statement or commitment to support for Python 3.13.

as for
> And you now have to re-litigate every one of those changes. I am not in favor of that.

The thing to re-litigate is deprecating `error_components` and `derivatives`. I would like to revert this because it is BROKEN in 3.2.4 AND I am having misgivings about our lack of migration plan to 4.0. Otherwise nothing in the 3.2.4 changelog needs re-litigation.

> Yes, Python warnings do not work well. The available choices are

I don't mind `FutureWarning` being emitted everytime the USER calls `error_components()` or `derivatives`. We are really trying to warn the user about these things. The problem is that our own internal calls to `error_components()` and `derivatives` are raising these warnings making it impossible for users on 3.2.4 to avoid getting these warnings, even if they are "behaving" and not calling `error_components()` explicitly themselves.

newville:
@jagerber48 Your proposal requires a PR.  That does not exist.  You can make that PR right now -- you could have made one at any point in this process.  Or you can approve this one AND THEN un-deprecate anything you want.     Code *can* be undeprecated after being warned as PendingDeprecationWarning -- that is actually a clear feature of pending deprecation warnings.

Move forward.  Do not ignore maintenance.

There is a very real danger of precipitating a non-friendly fork here.

newville:
@jagerber48 I will enable "merge without waiting for requirements to be met" (ie, without an approved review) and merge this branch in 2 days if it is not approved. 

And then we will need a very real conversation about how this project is managed.    This is not good.

neutrinoceros:
@newville I genuinely don't understand why you're pressuring your peers into approving this PR, which again, doesn't address the problems it's claiming to. As a third party consumer, your attitude is a bit alarming.

jagerber48:
@newville

- Yanking 3.2.4 mitigates the danger of an unfriendly fork.
- I have issues with this PR. I think even in this "emergency" situation, it is the wrong move.
- I like #354 PR. This, short-term, resolves the issues people are experiencing. I have been moving forward with it and think we could have it merged by EOD. It exactly solves the problems users are experiencing and doesn't introduce further user-facing problems that we need to deal with in further releases. It leaves us with internal details to figure out but that is ok. It gets us out of our "urgent" emergency.

Note, this is the purpose of the yank feature on pypi. If you make a mistake and push a bad release then, if the team makes the decision, you can have the release quickly yanked within 5 minutes so users don't face issues while you sort through things internally.

jagerber48:
> @newville I genuinely don't understand why you're pressuring your peers into approving this PR, which again, doesn't address the problems it's claiming to. As a third party consumer, your attitude is a bit alarming.

@newville I agree with this. Your attitude has pushed away the other people who volunteered to help maintain this repo when lebigot made the call. Your attitude also pushes away friendly folks in the community offering their time to help make `uncertainties` better. I am concerned that *this* fork has become the "unfriendly" fork.

newville:
@jagerber48 This PR is to fix the FutureWarning.  I did not add that code, I did not push that release.  In fact, my PR for 3.2.4 was *ignored*.    This is just "let's turn those messages off and go back to figuring it out".   You can do anything you want after that.  You are delaying fixing the actual problem.  

But, OK, I'm the bad guy.   

OK.  I yanked 3.2.4.  

Closing this PR.   That's 2 PRs in a row not merged.  yep, good luck.

PR #354, opened 2026-01-22
Fix internal usage of deprecated APIs triggering user warnings

### LINKED ISSUE #352: DEPR: self-triggered deprecation warnings in 3.2.4

Littie28:
Added _derivatives and _error_components for internal use. Public APIs retain deprecation warnings as intended.

Fixes #352

- [x] Closes #352 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Change Log
 ===================
 
+3.2.4a  2026-January-22
+-----------------------
+
+Adds:
+- Add new internal _derivatives and _error_components properties for internal use without deprecation warning
+
 3.2.4   2026-January-9
 -----------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Change Log
 ===================
 
-3.2.4a  2026-January-22
+Unreleased
 -----------------------
 
 Adds:
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -433,15 +433,7 @@ class AffineScalarFunc(object):
     @property
     def _derivatives(self):
         """
-        Return a mapping from each Variable object on which the function
-        (self) depends to the value of the derivative with respect to
-        that variable.
-
-        This mapping should not be modified.
-
-        Derivative values are always floats.
-
-        This mapping is cached, for subsequent calls.
+        Private wrapper for derivatives without raising a FutureWarning.
         """
         if not self._linear_part.expanded():
             self._linear_part.expand()
@@ -456,7 +448,15 @@ class AffineScalarFunc(object):
     @property
     def derivatives(self):
         """
-        Public wrapper for _derivatives.
+        Return a mapping from each Variable object on which the function
+        (self) depends to the value of the derivative with respect to
+        that variable.
+
+        This mapping should not be modified.
+
+        Derivative values are always floats.
+
+        This mapping is cached, for subsequent calls.
         """
         warn(
             f"{self.__class__.__name__}.derivatives() is deprecated. It will "
@@ -471,7 +471,14 @@ class AffineScalarFunc(object):
     # Uncertainties handling:
     def error_components(self):
         """
-        Publich wrapper of _error_compents with FutureWarning.
+        Individual components of the standard deviation of the affine
+        function (in absolute value), returned as a dictionary with
+        Variable objects as keys. The returned variables are the
+        independent variables that the affine function depends on.
+
+        This method assumes that the derivatives contained in the
+        object take scalar values (and are not a tuple, like what
+        math.frexp() returns, for instance).
         """
         warn(
             f"{self.__class__.__name__}.error_components() is currently an "
@@ -487,14 +494,7 @@ class AffineScalarFunc(object):
 
     def _error_components(self):
         """
-        Individual components of the standard deviation of the affine
-        function (in absolute value), returned as a dictionary with
-        Variable objects as keys. The returned variables are the
-        independent variables that the affine function depends on.
-
-        This method assumes that the derivatives contained in the
-        object take scalar values (and are not a tuple, like what
-        math.frexp() returns, for instance).
+        Private wrapper for error_compents without FutureWarning.
         """
         # Calculation of the variance:
         error_components = {}
--- a/uncertainties/core.py
+++ b/uncertainties/core.py
@@ -430,10 +430,8 @@ class AffineScalarFunc(object):
 
     ############################################################
 
-    # Making derivatives a property gives the user a clean syntax,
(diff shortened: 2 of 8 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Change Log
 ===================
 
+3.2.4a  2026-January-22
```
jagerber48: Please change this label to "Unreleased". The `uncertainties` maintainers will change the label appropriately (probably to `3.2.5`) in our release workflow.

uncertainties/core.py
```diff
@@ -460,20 +451,27 @@ def derivatives(self):
 
         return self._linear_part.linear_combo
 
+    # Making derivatives a property gives the user a clean syntax,
+    # which is consistent with derivatives becoming a dictionary.
+    @property
+    def derivatives(self):
+        """
+        Public wrapper for _derivatives.
```
jagerber48: Please keep the historic docstring on the public, not private `derivatives` property.

uncertainties/core.py
```diff
@@ -460,20 +451,27 @@ def derivatives(self):
...
         """
-        Individual components of the standard deviation of the affine
-        function (in absolute value), returned as a dictionary with
-        Variable objects as keys. The returned variables are the
-        independent variables that the affine function depends on.
-
-        This method assumes that the derivatives contained in the
-        object take scalar values (and are not a tuple, like what
-        math.frexp() returns, for instance).
+        Publich wrapper of _error_compents with FutureWarning.
```
jagerber48: Please keep the historic docstring on the public, not private, `error_components()` method.

jagerber48 (commented):
@Littie28 thank you very much for this contribution!

Please make sure you've changed all internal usages of `.derivatives` and `.error_components()` to use the private versions. Our CI won't catch cases that you've missed under it's normal configuration at the moment.

uncertainties/core.py
Littie28: Moved historic docstring to public api.

uncertainties/core.py
Littie28: Moved historic docstring to public api.

CHANGES.rst
Littie28: Updated CHANGES.rst

Littie28:
I searched through the codebase for usages of `.derivatives` and `.error_components()`. They currently appear only in comments and tests. Should the test suite use the private versions (no FutureWarning) or the public API (with FutureWarning)?

jagerber48:
> I searched through the codebase for usages of `.derivatives` and `.error_components()`. They currently appear only in comments and tests. Should the test suite use the private versions (no FutureWarning) or the public API (with FutureWarning)?

Thanks for that! Let’s use the private versions in the tests. I want to change tests to error on warnings but I don’t want error_components usage to cause tests to fail. In the future I’m interested to move away from derivatives and error components in tests. This will be a reminder for that.

uncertainties/core.py
```diff
@@ -14,24 +14,24 @@
 
 from __future__ import division  # Many analytical derivatives depend on this
 
-from builtins import str, zip, range, object
+import collections
```
jagerber48: Please don't reorganize imports. You might have some local automation that is doing that but we don't want these kinds of changes to pollute this PR. I don't disagree with doing this sort of thing, but this PR is not the place.

uncertainties/core.py
```diff
@@ -430,6 +430,19 @@ def nominal_value(self):
 
     ############################################################
 
+    @property
+    def _derivatives(self):
+        """
+        Private wrapper for _derivatives without raising a FutureWarning.
```
jagerber48:
```
"""
Private version of `derivatives` for internal use.
"""
```

The FutureWarning thing is kind of a temporary fleeting issue. Not worth citing in the docstring. Sorry for being picky about wording.

uncertainties/core.py
```diff
@@ -485,10 +490,16 @@ def error_components(self):
             stacklevel=2,
         )
 
+        return self._error_components()
+
+    def _error_components(self):
+        """
+        Private wrapper for _error_compents without FutureWarning.
```
jagerber48:
```
"""
Private version of `error_components` for internal use.
"""
```

The FutureWarning thing is kind of a temporary fleeting issue. Not worth citing in the docstring. Sorry for being picky about wording.

uncertainties/core.py
```diff
@@ -580,7 +591,7 @@ def std_score(self, value):
             # here:
             return (value - self._nominal_value) / self.std_dev
         except ZeroDivisionError:
-            raise ValueError("The standard deviation is zero:" " undefined result")
+            raise ValueError("The standard deviation is zero: undefined result")
```
jagerber48: Again, this change is ok but this is not the right place to address it. It pollutes the PR.

Littie28:
Rebasing to remove the style/formatting changes.

uncertainties/core.py
Littie28: All formatting changes are removed

uncertainties/core.py
Littie28: All formatting changes are removed

uncertainties/core.py
Littie28: Updated docstrings

uncertainties/core.py
Littie28: Updated docstrings

jagerber48:
Stepping away from `uncertainties` for a few days. I think it is likely this PR will still be merged into `master` as the start of the `3.2.5` release which will re-release everything that was in `3.2.4` but resolve the "internal calls to `error_components()` or `derivatives` spamming the users with warnings.

Littie28:
Thanks for the guidance throughout! No rush at all. I'll keep an eye on the PR and address any feedback when you're back. Enjoy the break!

PR #358, opened 2026-06-17
`ufloat_fromstr` can now also parse uncertainties as percentages

### LINKED ISSUE #357: uncertainties as percentages
Hi there,

Working with this package I found that it can not read strings with percentage uncertainties, such as:
> 23 +- 10%
> 345 +/- 2%
> 432 (+- 34%)
> etc.

I ended up including it in the `parsing.py`, so that these numbers can be read with the `uncertainties.ufloat_fromstr`. I will open a pull request in case it is useful.

pablogila:
- [x] Closes #357

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Adds:
 - Adjusts the Sphinx configuration to allow for reproducible builds using
   ``SOURCE_DATE_EPOCH``.
 - Mark Python 3.13 and 3.14 as officially supported.
+- Support for percentage uncertainties in the `ufloat_fromstr` function.
 
 Deprecates:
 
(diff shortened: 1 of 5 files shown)


newville (approved):

newville:
@pablogila Sorry for the delay. As at #357, this isn't a feature I would normally use, but it looks OK to me, and if others want to use this, I am OK with merging. I might say we should wait a bit to give others a chance to review this, but it has been two weeks, so maybe we can plan to merge in a few days unless there are concerns from others.

jagerber48:
@pablogila Thank you for this contribution!

This looks good to me for merging. I don't have bandwidth to comb through the regexp and all the logic but the tests look good and old tests don't break.

To address some comments from the linked issue: I won't comment as to how much use I think this feature will get, but I'll say that I think it's fine to include a feature like this given `uncertainties` supports some amount of round tripping between ufloat objects and formatted strings. In other words, when I saw that `uncertainties` can parse strings its creates back into ufloat objects, my expectation was that it should be able to convert ALL formatted strings it generates back into ufloat objects and this PR helps support that expectation. 

So this PR has a thumbs up for merging from me!

jagerber48 (approved):

newville:
@jagerber48 Thanks - I completely agree with all of that.  It's working for someone, might be useful for others, doesn't appear to break anything, doesn't look like a maintenance burden.   Merging.

PR #359, opened 2026-07-04
Fix umath.frexp() uncertainty propagation onto the mantissa

gaoflow:
`umath.frexp()` drops the uncertainty on the returned mantissa and raises when it is read:

```python
>>> from uncertainties import ufloat
>>> from uncertainties.umath import frexp
>>> frexp(ufloat(3.0, 0.1))[0].std_dev
TypeError: cannot unpack non-iterable LinearCombination object
```

The mantissa's linear part was built as `LinearCombination([2**-exponent, aff_func._linear_part])`, i.e. a bare two-element list, but `LinearCombination.expand()` unpacks each element as a `(coefficient, term)` pair. The nominal value is fine, but the crash surfaces lazily when the uncertainty is expanded. The sibling `ldexp` (a few lines above) already uses the correct pair form `[(coeff, part)]`.

Fix wraps the single term in a tuple, matching `ldexp` and `expand()`'s contract:

```python
LinearCombination([(2**-exponent, aff_func._linear_part)]),
```

`test_frexp` checks the propagated `std_dev` equals the analytic `2**-e * sigma` (including negative inputs, exact powers of two at the octave boundary, and a multi-variable combination that exercises the previously-crashing `expand` path). Full `test_umath.py` passes (286), `ruff check`/`format` clean.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,12 @@ Adds:
 - Mark Python 3.13 and 3.14 as officially supported.
 - Support for percentage uncertainties in the `ufloat_fromstr` function.
 
+Fixes:
+
+- Fixes `umath.frexp()` so that the uncertainty is correctly propagated onto the
+  returned mantissa. Previously, reading the `std_dev` of the mantissa raised a
+  `TypeError`.
+
 Deprecates:
 
 - The `AffineScalarFunc.derivatives` property has been marked as deprecated. This
--- a/tests/test_umath.py
+++ b/tests/test_umath.py
@@ -282,6 +282,88 @@ def test_hypot():
     assert isnan(result.derivatives[y])
 
 
+def test_frexp():
+    """
+    frexp() must propagate the uncertainty onto the returned mantissa.
+
+    For frexp(x) = (m, e), we have x = m * 2**e with e a
+    locally-constant integer, hence m = x * 2**-e and dm/dx = 2**-e.
+    This is a regression test for a malformed ``LinearCombination`` in
+    the mantissa: it was built as ``[coeff, linear_part]`` instead of
+    ``[(coeff, linear_part)]``, so merely reading the mantissa's
+    ``std_dev`` raised ``TypeError: cannot unpack non-iterable
+    LinearCombination object``.
+    """
+    cases = [
+        (3.0, 0.1),
+        (-12.5, 0.4),
+        (1.0, 0.2),
+        (0.5, 0.05),
+        (1024.0, 8.0),
+        (-0.03125, 1e-3),
+    ]
+    for nominal, std_dev in cases:
+        x = ufloat(nominal, std_dev)
+        with pytest.warns(FutureWarning, match="will be removed"):
+            mantissa, exponent = umath_core.frexp(x)
+
+        ref_mantissa, ref_exponent = math.frexp(nominal)
+        # Nominal value and exponent match the plain-float result:
+        assert mantissa.nominal_value == ref_mantissa
+        assert exponent == ref_exponent
+        # The exponent is a plain int and carries no uncertainty (it is
+        # returned as-is, not wrapped into an AffineScalarFunc):
+        assert isinstance(exponent, int)
+        assert not isinstance(exponent, uncert_core.AffineScalarFunc)
+        # frexp postcondition on the returned mantissa magnitude:
+        assert 0.5 <= abs(mantissa.nominal_value) < 1.0
+
+        # Reading the uncertainty must not raise, and must equal the
+        # analytical first-order propagation dm/dx = 2**-e:
+        assert nan_close(mantissa.std_dev, 2.0**-exponent * std_dev)
+
+        # The mantissa is an exact linear function of x, so
+        # m * 2**e is fully correlated with x and their difference
+        # carries neither a nominal value nor an uncertainty:
+        residual = mantissa * 2.0**exponent - x
+        assert residual.nominal_value == 0
+        assert residual.std_dev == 0
+
+    # A mantissa whose linear part spans several variables exercises the
+    # full expansion of the (previously malformed) linear combination:
+    a = ufloat(5.0, 0.2)
+    b = ufloat(3.0, 0.1)
+    with pytest.warns(FutureWarning, match="will be removed"):
+        mantissa, exponent = umath_core.frexp(a + b)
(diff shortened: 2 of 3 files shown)


newville:
@gaoflow I have to admit that my first reaction is "wait, why is `frexp()` supported in `uncertainties.umath`?"    And, indeed it is deprecated, which is sensible and comes as a relief (at least to me).

I can see that `math.frexp(uval.n)` and `math.frexp(uval.s)`  may be interesting, but I think `math.frexp(uval)` may not really make any sense.  Your tests all have `abs(uval.s) < abs(uval.n)`.  But  `u = 0.6 +/- 3.0` is a sensible `ufloat`, and 

   frexp(0.6) =  (0.6, 0) 
   frexp(3.6) =  (0.9, 2)
   frexp(-2.4) = (-0.6, 2)

What is the result supposed to be other than ( (0.6 +/- 3.0), 0) ?   That's kind of not a valid value for `frexp()`, but I think one is not possible.  

Sorry, but it is hard to be excited about fixing deprecated functions.

binggao1230:
Fair — it's deprecated and I won't argue it deserves much attention. One clarification on what the PR actually changes, though: on `main`, `frexp(x)` for an `AffineScalarFunc` doesn't just return a questionable value, it builds a malformed derivative and raises `TypeError` the moment you read `.std_dev` on the mantissa. So this replaces a crash rather than adding support for something new.

And the result is exactly the `((0.6 +/- 3.0), 0)` you wrote: the exponent comes from the nominal value (as with every other branch-selecting function here) and the mantissa carries `s / 2**e` by the usual first-order rule. A std_dev that overflows `[0.5, 1)` is just honest linear propagation — the same thing happens for any function evaluated where the uncertainty is large relative to curvature.

That said, I agree the payoff is small for a deprecated function. If you'd rather not carry it, I'm happy to close — no objection.

newville:
@gaoflow I sort of don't understand how to make sense of a result from `frexp` that is (0.6 +/-3.0, 0).

The mantissa from `frexp` needs to be on [0.5, 1.0], otherwise the power of 2 has to change.  I see a use-case for `math.frexp(uval.n)` and `math.frexp(uval.s)`, but kind of not for `umath.frexp(uval)`

But maybe I'm not understanding something (that's been known to happen... basically daily ;)). 

Anyone else got any thoughts or an opinion on this?

binggao1230:
I think the reconciling point is that here `x +/- s` isn't an interval — it's a nominal value plus a first-order (linearized) 1σ. frexp's *exponent* is a function of the nominal only (`frexp(0.6)` → exponent 0, mantissa 0.6 ∈ [0.5, 1)), and the mantissa then carries the propagated `s / 2**e`. So the `3.0` isn't saying the mantissa reaches 3.6 — it's the linear sensitivity, the same way `sin(x)` can have σ > 1 for a large input σ without anything being out of range.

Concretely it's just the inverse of `ldexp`, which umath already propagates exactly this way: `ldexp(m, e)` scales the mantissa's uncertainty by `2**e`, so `frexp` scales by `2**-e`. Same model, opposite direction.

And the practical choice isn't support vs. no support — today `frexp(uval)` is already wired up but *crashes* (TypeError on `.std_dev`). So it's crash vs. the first-order result: `frexp(uval.n)` gives you the exponent, and `frexp(uval)` gives that same exponent plus the mantissa's propagated uncertainty in one call.

Still totally fine if you'd rather not carry a deprecated function, though — just wanted the model to be clear.

newville:
@gaoflow 
> I think the reconciling point is that here x +/- s isn't an interval — it's a nominal value plus a first-order (linearized) 1σ. 

Yeah, I get that.  My point was meant to be that when a number can have a range of values that exceeds a factor of 2, `frexp` is not so obvious.  For reference, numpy's `frexp` returns an ndarray of mantissas and an ndarray of exponents.   Each number is broken down separately.  That kind of doesn't work here.  

I think I would have the same confusion about `modf` and `divmod` -- probably and function that decomposes a float into two values:  that is clear for a float, I don't see an obvious mapping to a ufloat.

Again, `math.frexp(ufloat.n)` and `math.frexp(ufloat.s)`can be useful and meaningful.  

Is there a sensible use-case for this function?  What does one do with the result of `umath.frexp()`? 

Trying to understand.  Hoping that someone else has some insight.

binggao1230:
Honestly, that's a fair challenge and I don't want to manufacture a use-case that isn't there. You're right that decomposition functions — `frexp`, `modf`, `divmod` — don't have a clean ufloat semantics: one of the two outputs is inherently discrete, and "which exponent" stops being well-defined once the uncertainty spans an octave. `frexp(uval.n)` / `frexp(uval.s)` cover the cases that actually make sense.

So I'd narrow the claim: the only concrete defect here is that `frexp(uval)` doesn't currently fall through to *anything* sensible — it raises an opaque `TypeError` from deep in the internals. This PR turns that into the first-order result, but I take your point that a first-order result for a semantically-shaky operation may not be what you want.

Given it's deprecated anyway, I'm happy to just close this. The one alternative, if avoiding the opaque crash has any value to you, would be to make it raise a clear "frexp doesn't support uncertainty propagation" error instead — but that may well be more churn than it's worth. Your call; no objection either way.

jagerber48:
Yes, `umath.frexp` and others were deprecated because they don't follow the right semantics for `ufloat` operations. Good functions are continuous functions, with non-zero derivative which accept a real number and return a real number. Functions with multiple inputs or output should not be supported by `uncertainties`, these are "strongly" bad functions. Functions like `x^2` are also bad at `x=0` but this is more of an edge case of badness which can be documented and handled by users.

A lot of the functions in `umath` came in as a part of an effort at one point in time to make `umath` an uncertainties aware version of `math`, having each function in `math` supported in `umath` with the idea to port every function over. We've since decided that is not a worthwhile goal for reasons exactly like this one.

See #298 and #295 for more information.

I don't think we should support development on this deprecated feature which I am eager to see dropped when (if ever) we release 4.0

@gaoflow Thanks for the PR and studying the data model. However, I'm going to go ahead and close this issue.

newville:
@jagerber48 Thanks -- I was willing to try to hear a use-case for these, but I think it is better to just stop supporting these.  I also understand @gaoflow's point that the current implementation simply fails, which is not good.

I'm definitely +1 on seeing version 4, with getting these deprecated functions fully removed being a small but important part of that.

